PR #98, opened 2020-11-13
Small refactor for cleanup

ekimd:
Break out token handling into separate methods for readability.

### DIFF
--- a/sqllineage/core.py
+++ b/sqllineage/core.py
@@ -79,9 +79,9 @@ class LineageAnalyzer:
         :param stmt: a SQL statement parsed by `sqlparse`
         """
         if stmt.get_type() == "DROP":
-            self._extract_from_DDL_DROP(stmt)
+            self._extract_from_ddl_drop(stmt)
         elif stmt.get_type() == "ALTER":
-            self._extract_from_DDL_ALTER(stmt)
+            self._extract_from_ddl_alter(stmt)
         elif (
             stmt.get_type() == "DELETE"
             or stmt.token_first(skip_cm=True).normalized == "TRUNCATE"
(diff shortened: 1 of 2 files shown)


sqllineage/core.py
```diff
@@ -91,113 +91,116 @@ def analyze(self, stmt: Statement) -> LineageResult:
...
+                continue
+
             if isinstance(sub_token, TokenList):
-                self._extract_from_DML(sub_token)
+                self._extract_from_dml(sub_token)
+
             if sub_token.ttype in Keyword:
-                if any(
-                    re.match(regex, sub_token.normalized)
-                    for regex in SOURCE_TABLE_TOKENS
```
reata, line 119: This part is following [black ](https://github.com/psf/black) code style. I'd like to stay with that.

reata:
Hi, ekimd. Thanks for this PR. I do like this idea of token handling logic breakdown. One thing that may not catch your attention is that I'm trying to enforce [black](https://github.com/psf/black#version-control-integration) code style in this repo. That's where this [.pre-commit-config.yaml](https://github.com/reata/sqllineage/blob/master/.pre-commit-config.yaml) is coming from.

So would you manually do a black check for the file you're changing? That would be really helpful. Thanks. (FYI, I should really add this black check in GitHub Actions so you would know. My bad.)

sqllineage/core.py
```diff
@@ -79,9 +79,9 @@ def analyze(self, stmt: Statement) -> LineageResult:
         :param stmt: a SQL statement parsed by `sqlparse`
         """
         if stmt.get_type() == "DROP":
-            self._extract_from_DDL_DROP(stmt)
```
reata, line 82: I hesitate for a long time that whether we should use upper case or lower case here. Since they're abbreviation, upper case should they be. But it's bad code style to use upper case for method name. Would you share your opinion?

reata (changes requested):
Stay with black code style. Otherwise it looks good to me.

sqllineage/core.py
ekimd: Understood.  I've pushed a commit with black output.

sqllineage/core.py
ekimd: Technically, uppercase method names violate PEP 8.  My standard code practice is to try to eliminate any lint warning messages.

reata (approved):
LGTM

sqllineage/core.py
reata: Sure, let's make it lower case

Issue #93, opened 2020-11-10
Incorrect Result for Cases When Non-Reserved Keyword Used as Table Name


coldmoutain:
hello,reata:
I got the newest code from your github to test whether SQLLINEAGE works.
it works very well for most of  cartesian product SQL.
 but ,when i input sql " select a.* from table a,table b" something strange happened:
### ` LineageRunner("select a.* from source a,source b").source_tables`  returns "[Table: <default>.a]" but not " table"
I run this sql just for fun ,after found this BUG .I tested more and get another "strange returns"
` LineageRunner("select a.* from source1 a,source b").source_tables` returns " [Table: <default>.source1]"
I thought , Was it the Tablenames matter? so I changed the Tablenames
`LineageRunner("select * from care a,care b").source_tables` returns "[Table: <default>.care]",It works.
`LineageRunner("select * from care1 a,care b").source_tables` returns "[Table: <default>.care, Table: <default>.care1]" It works well.
### `LineageRunner("select * from care a,care b").source_tables` returns "[Table: <default>.care]" but not "a" 

well,I thought that maybe some Tablenames sqllineage can`t works well.

reata:
Thanks for reporting this. It's interesting. Let me check.

reata:
```python
>>> from sqllineage.runner import LineageRunner
>>> print(LineageRunner("select a.* from source a,source b").source_tables)
[Table: <default>.a]
>>> print(LineageRunner("select a.* from source1 a,source b").source_tables)
[Table: <default>.source1]
>>> print(LineageRunner("select * from care a,care b").source_tables)
[Table: <default>.care]
>>> print(LineageRunner("select * from care1 a,care b").source_tables)
[Table: <default>.care, Table: <default>.care1]
```

The first two lineage are wrong. No. 1 should be [Table: <default>.source] and No. 2 should be [Table: <default>.source, Table: <default>.source1]. And the last two lineage are correct.

Actually You're unlucky to hit a corner case again. This issue arises from the fact that `source` is a keyword in SQL language, even though it's not reserved keyword like `select` or `insert`). See https://dev.mysql.com/doc/refman/8.0/en/keywords.html for reference.

It's definitely a bug. However, it will take some time for an elegant solution since I can't distinguish reserved keyword or any regular keyword right now.

reata:
As a short term solution, you can try escape the table name source with back tick in MySQL:
```python
>>> print(LineageRunner("select a.* from `source` a,`source` b").source_tables)
[...]

PR #104, opened 2020-12-06
refactor: remove dependency on sqlparse.tokens module

reata:
Preparation work for #93 
Remove all dependency on `sqlparse.tokens` module, make it unified to depend on `sqlparse.sql` module.


reata:
Since we're removing redundant code and got 100% diff hit, it's OK to get absolute coverage drop

PR #114, opened 2021-01-21
Parsing of multiple with statments added

Vendetta01:
The following SQL statement results in an Exception:
```sql
WITH wtab1 AS (SELECT * FROM tab1),
wtab2 AS (SELECT * FROM tab2)
SELECT *
FROM wtab1
```

The reason is that after a with statement an _Identifier_ is assumed. The above example however results in an _IdentifierList_. This pull request addresses this issue by extending *_handle_temp_table_token* to also check for *IdentifierList* and resolving that case as needed.
I also added two new tests.

Last but not least let me thank you for your great work on this project!

### DIFF
--- a/sqllineage/core.py
+++ b/sqllineage/core.py
@@ -202,18 +202,16 @@ class LineageAnalyzer:
 
     def _handle_temp_table_token(self, sub_token: TokenList) -> None:
         if isinstance(sub_token, Identifier):
-            self._lineage_result.intermediate.add(
-                Table.create(sub_token))
+            self._lineage_result.intermediate.add(Table.create(sub_token))
             self._extract_from_dml(sub_token)
         elif isinstance(sub_token, IdentifierList):
             for temp_tab_token in sub_token:
                 if isinstance(temp_tab_token, Identifier):
-                    self._lineage_result.intermediate.add(
-                        Table.create(temp_tab_token))
+                    self._lineage_result.intermediate.add(Table.create(temp_tab_token))
                     self._extract_from_dml(temp_tab_token)
         else:
             raise SQLLineageException(
-                "An Identifier(List) is expected, got %s[value: %s] instead"
+                "An Identifier or IdentifierList is expected, got %s[value: %s] instead"
                 % (type(sub_token).__name__, sub_token)
             )
 
--- a/sqllineage/core.py
+++ b/sqllineage/core.py
@@ -201,13 +201,21 @@ class LineageAnalyzer:
                 self._lineage_result.write.add(Table.create(sub_token))
 
     def _handle_temp_table_token(self, sub_token: TokenList) -> None:
-        if not isinstance(sub_token, Identifier):
+        if isinstance(sub_token, Identifier):
+            self._lineage_result.intermediate.add(
+                Table.create(sub_token))
+            self._extract_from_dml(sub_token)
+        elif isinstance(sub_token, IdentifierList):
+            for temp_tab_token in sub_token:
+                if isinstance(temp_tab_token, Identifier):
+                    self._lineage_result.intermediate.add(
+                        Table.create(temp_tab_token))
+                    self._extract_from_dml(temp_tab_token)
+        else:
             raise SQLLineageException(
-                "An Identifier is expected, got %s[value: %s] instead"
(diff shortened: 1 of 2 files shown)


tests/test_select.py
```diff
@@ -142,3 +142,17 @@ def test_select_join_in_ansi89_syntax():
...
+        "WITH wtab1 AS (SELECT * FROM schema1.tab1) SELECT * FROM wtab1",
+        {"schema1.tab1"})
+
+def test_with_select_many():
+    helper(
+        "WITH wtab1 AS (SELECT * FROM schema1.tab1), "
+        "wtab2 AS (SELECT * FROM schema2.tab2)"
+        "SELECT * FROM wtab1 JOIN wtab2 ON wtab1.id = wtab2.id",
+        {"schema1.tab1", "schema2.tab2"}
+    )
```
reata, line 158: 1. Change the testing SQL to the following. With column name specified, I believe it makes more sense for multiple CTE subclauses. FYI, I get this SQL from MySQL manual (https://dev.mysql.com/doc/refman/8.0/en/with.html). 
```sql
WITH
  cte1 AS (SELECT a, b FROM table1),
  cte2 AS (SELECT c, d FROM table2)
SELECT b, d FROM cte1 JOIN cte2
WHERE cte1.a = cte2.c;
```

2. Add a new line at end of file. It's always good coding style to end a file with newline.

sqllineage/core.py
```diff
@@ -201,13 +201,21 @@ def _handle_target_table_token(self, sub_token: TokenList) -> None:
                 self._lineage_result.write.add(Table.create(sub_token))
 
     def _handle_temp_table_token(self, sub_token: TokenList) -> None:
-        if not isinstance(sub_token, Identifier):
+        if isinstance(sub_token, Identifier):
+            self._lineage_result.intermediate.add(
+                Table.create(sub_token))
```
reata, line 206: I'm trying to enforce black code style. So according to black check, make this one line. 

You can see the black check result in GitHub actions (https://github.com/reata/sqllineage/pull/114/checks?check_run_id=1744343710) for this PR.
Or better, you can try adding [pre-commit](https://pre-commit.com/) hoot for your local git repo so each time before git commit, some sort of checks will be enforced before pushing to GitHub. I have a [pre-commit configuration file](https://github.com/reata/sqllineage/blob/master/.pre-commit-config.yaml) in the repo.

sqllineage/core.py
```diff
@@ -201,13 +201,21 @@ def _handle_target_table_token(self, sub_token: TokenList) -> None:
...
+            self._extract_from_dml(sub_token)
+        elif isinstance(sub_token, IdentifierList):
+            for temp_tab_token in sub_token:
+                if isinstance(temp_tab_token, Identifier):
+                    self._lineage_result.intermediate.add(
+                        Table.create(temp_tab_token))
```
reata, line 212: Same as above, make it one line.

sqllineage/core.py
```diff
@@ -201,13 +201,21 @@ def _handle_target_table_token(self, sub_token: TokenList) -> None:
...
+                    self._extract_from_dml(temp_tab_token)
+        else:
             raise SQLLineageException(
-                "An Identifier is expected, got %s[value: %s] instead"
+                "An Identifier(List) is expected, got %s[value: %s] instead"
```
reata, line 216: We might as well make it explicit here. After all, [Explicit is better than implicit](https://www.python.org/dev/peps/pep-0020/).

"An Identifier or IdentifierList is expected, got %s[value: %s] instead"

reata (changes requested):
Thanks for your PR. Personally I rarely use CTE, and it's good to get cases like this covered.

The overall implementation looks good to me, except for some coding style issues. I made some comments, please change the code accordingly. 

And please fix all the broken GitHub Actions checks before next review. I know these're hidden rules running background that may seem unclear to you. I'll add a contributing guide to this repo later. Before that, just check GitHub Actions, see the log and follow what it says. Sorry for the inconvenience.

Vendetta01:
Thank you for your detailed feedback, I incorporated all requested changes into the branch.

reata (approved):
LGTM

PR #137, opened 2021-04-23
Option to specify hostname for flask app that spins up for graph visualization

surajpotnuru:


sqllineage/drawing.py
```diff
@@ -80,6 +80,7 @@ def find_children(folder: Path):
 
 def draw_lineage_graph(**kwargs) -> None:
     port = kwargs.pop("p", DEFAULT_PORT)
+    ip = kwargs.pop("i", DEFAULT_HOST)
     querystring = urlencode({k: v for k, v in kwargs.items() if v})
-    print(f" * SQLLineage Running on http://localhost:{port}/?{querystring}")
-    app.run(port=port)
+    print(f" * SQLLineage Running on http://{ip}:{port}/?{querystring}")
+    app.run(host = ip, port=port)
```
reata, line 86: Please remove the space between the assign operator

reata (changes requested):
The sqllineage.cli module should also be updated, or we won't be able to specify the host option. And I could prefer to use -h --host

reata:
@surajpotnuru Since I haven't heard from you for a while, meanwhile I'm working on a sqllineage demo deployment, which also requires this feature. So I submit a PR for this. Specifying hostname option is now available in v1.2.1. Thanks for your interest in this project anyway.

Issue #164, opened 2021-09-01
Support for Redshift 'copy from' syntax


RickardCardell:
I would like to add some missing Redshift syntax support, namely 'COPY target FROM src'.
The syntax specification: https://docs.aws.amazon.com/redshift/latest/dg/r_COPY.html

The biggest change is that would allow a string literal as the source, e.g `'s3://path/'`:
`COPY target FROM 's3://some_path'`

It can be discussed if having a source table being represented as a path is something that should be supported.

reata:
I checked the PR and code is fine. As you said, the only thing that should be discussed is whether we'd want a file path (be it s3, hdfs, or just local file) as table.

In Spark world which I'm familiar with, we can also do things like "SELECT * FROM parquet.\`examples/src/main/resources/users.parquet\`" (https://spark.apache.org/docs/latest/sql-data-sources-load-save-functions.html#run-sql-on-files-directly), though it's rarely used. Creating a table pointing to a path via hive metastore is more of a natural thing to do. Anyway we don't officially support this but it won't throw exception either.
```bash
$ python -m sqllineage.cli -e 'insert into tab1 select * from parquet.`examples/src/main/resources/users.parquet`'
Statements(#): 1
Source Tables:
    examples/src/main/resources/users.parquet
Target Tables:
    <default>.tab1
```
Here examples/src/main/resources/users is identified as database while the trailing parquet identified as table, with heading parquet being ignored. Not so good even though the result looks fine. The ideal way for me is to create another Path data model and not mixed it with Table if we really want this. But that's more of a long term thing.

I guess in redshift world, "COPY tab1 FROM tab2" is far less used than "COPY tab1 FROM 's3://bucket/file/path'" and loading data from S3 is a regular thing to do? If not, I would prefer the string literal support being removed. Would you share some of your thought here?

RickardCardell:
Thanks for your time

> The ideal way for me is to create another Path data model and not mixed it with Table if we really want this. But that's more of a long term thing.

+1, I totally agree

> I guess in redshift world, "COPY tab1 FROM tab2" is far less used than "COPY tab1 FROM 's3://bucket/file/path'" and loading data from S3 is a regular thing to do? If not, I would prefer the string literal support being removed. Would you share some of your thought here?

[...]

PR #165, opened 2021-09-01
feature: add support for COPY FROM-syntax from the Redshift SQL dialect (#164)

RickardCardell:
For further details, see #164


reata:
Check #164 for the discussion.

Issue #172, opened 2021-10-15
Column Lineage Handling For Nested Token and Inline Comment


eeroel:
Hello,

First of all thank you @reata for this cool library! I was experimenting with a large set of SQL scripts and found a few cases for which the column extraction fails. The failures are mainly due to two issues, as far as I can see:
1. improper handling of nested tokens when extracting column names and
2. statements with window functions not handled properly

I'll paste the test cases here:
```
def test_comment_after_column():
    sql = """INSERT OVERWRITE TABLE tab1
    SELECT
    a.col1
    --, a.col2
    , a.col3
    FROM
    tab2 a
    """
    assert_column_lineage_equal(
        sql, [("tab2.col1", "tab1.col1"), ("tab2.col3", "tab1.col3")]
    )


def test_cast_with_comparison():
    sql = """
    INSERT OVERWRITE TABLE tab1
    select cast(col1=1 as int) col1, col2=col3 col2
    from tab2;
    """
    assert_column_lineage_equal(
        sql, [("tab2.col1", "tab1.col1"), ("tab2.col2", "tab1.col2"), ("tab2.col3", "tab1.col2")]
    )


def test_complex_window_function():
    sql = """
    INSERT OVERWRITE TABLE tab1
    SELECT tab2.col1 AS col1
        , MAX(tab2.col1) OVER (PARTITION BY tab2.col2) AS col2
    FROM tab2
    """
    assert_column_lineage_equal(
        sql, [("tab2.col1", "tab1.col1"), ("tab2.col1", "tab1.col2"), ("tab2.col2", "tab1.col2")]
    )

def test_window_function_in_subquery():
    sql = """
    insert into tab1
    select rn from (
        select
            ROW_NUMBER() OVER (PARTITION BY col1, col2) rn
        from tab2
    ) sub
    where rn = 1
    ;
    """
    assert_column_lineage_equal(
        sql, [("tab2.col1", "tab1.rn"), ("tab2.col2", "tab1.rn")]
    )
```

I have ideas how to fix some of these so I'll open a PR to demonstrate.

reata:
Thank you so much for submitting this issue as well as PR. I was also aware of both case, just hasn't decided which way to go yet.

1. Nested tokens
I observe column lineage parsing failure for complicated CASE clause. Token after keyword "WHEN" and keyword "THEN" can be a lot of things while I only covered the most simple case right now. 
[...]

PR #173, opened 2021-10-15
Recursive source table extraction, comment removal

eeroel:
- Call `Column._extract_source_raw_names` recursively to handle nested tokens
- Handle `IdentifierList` in `Column._extract_source_raw_names`
- Change string literals in `test_select_column_using_case_when` to use single quotes
https://github.com/reata/sqllineage/issues/172

### DIFF
--- a/sqllineage/models.py
+++ b/sqllineage/models.py
@@ -198,7 +198,6 @@ class Column:
                     # alias without AS
                     kw_idx, _ = token.token_next_by(i=Identifier)
                 idx, _ = token.token_prev(kw_idx, skip_cm=True)
-
                 expr = grouping.group(TokenList(token.tokens[: idx + 1]))[0]
                 source_raw_names = Column._extract_source_raw_names(expr)
                 return Column(alias, source_raw_names=source_raw_names)


tests/test_columns.py
```diff
@@ -75,19 +75,19 @@ def test_select_column_using_expression():
 
 def test_select_column_using_case_when():
     sql = """INSERT OVERWRITE TABLE tab1
-SELECT CASE WHEN col1 = 1 THEN "V1" WHEN col1 = 2 THEN "V2" END
+SELECT CASE WHEN col1 = 1 THEN 'V1' WHEN col1 = 2 THEN 'V2' END
```
eeroel, line 78: Note this change. Is it OK or is there a reason to consider double quoted strings as string literals here? This behavior is not consistent between RDBMS's, but being used to Postgres, to me "V1" and "V2" would intuitively correspond to column names instead of literals

sqllineage/models.py
```diff
@@ -197,6 +198,14 @@ def of(token: Token):
                     # alias without AS
                     kw_idx, _ = token.token_next_by(i=Identifier)
                 idx, _ = token.token_prev(kw_idx, skip_cm=True)
+                if idx is None:
```
eeroel, line 201: This is also a bit hacky

sqllineage/holders.py
```diff
@@ -98,7 +98,8 @@ def alias_mapping(self) -> Dict[str, Union[Table, SubQuery]]:
 
     def add_column_lineage(self, src: Column, tgt: Column) -> None:
         self.graph.add_edge(src, tgt, type=EdgeType.LINEAGE)
-        self.graph.add_edge(tgt.parent, tgt, type=EdgeType.HAS_COLUMN)
+        if tgt.parent is not None:
```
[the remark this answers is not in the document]
reata: src.parent could be None because of inaccurate column-to-table resolution. That's by design. But I don't see why tgt.parent could be None.

tests/test_columns.py
```diff
@@ -184,3 +184,55 @@ def test_select_column_from_same_table_multiple_time_using_different_alias():
...
+    INSERT OVERWRITE TABLE tab1
+    select cast(col1=1 as int) col1, col2=col3 col2
+    from tab2;
+    """
+    assert_column_lineage_equal(
+        sql, [("tab2.col1", "tab1.col1"), ("tab2.col2", "tab1.col2"), ("tab2.col3", "tab1.col2")]
+    )
+
+
+def test_complex_window_function():
```
reata, line 214: This is the only WIP part, right?

reata (changes requested):
1. Please fix the failed GitHub Action first. You can refer to https://github.com/reata/sqllineage/blob/master/CONTRIBUTING.md . Currently it's failing on the black code format check.
2. package-lock.json should be kept in the repo for a consistent build. See https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json
3. The overall proposed change looks good to me. I give my comment on some code block. Please take action accordingly.

sqllineage/models.py
eeroel: Yeah looks like this is due to https://github.com/andialbrecht/sqlparse/issues/394 . I fixed both cases now by running `sqlparse.format` with `strip_comments=True` as a preprocessing step in `runner`. Do you see any issue with this?

sqllineage/holders.py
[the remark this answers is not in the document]
eeroel: 👍 removed that

tests/test_columns.py
eeroel: Yes, and I decided to drop it from this PR because it clearly needs to be tackled in a different way

sqllineage/models.py
reata: That's a good one!

sqllineage/models.py
```diff
@@ -197,6 +198,7 @@ def of(token: Token):
                     # alias without AS
                     kw_idx, _ = token.token_next_by(i=Identifier)
                 idx, _ = token.token_prev(kw_idx, skip_cm=True)
+
```
reata, line 201: Can we get this empty line removed?

PR #245, opened 2022-04-06
initialize logging config at main

tozka:
Moving to initialize logging config at entry point of the CLI . This way
if used as a library (imported in another projected) it will not
override the other project logging configuration

Testing Done: Ran tox -e py310 and used pre-commit hook installed
locally.

Signed-off-by: Antoni Ivanov <aivanov@vmware.com>


reata (approved):
LGTM

Issue #257, opened 2022-05-08
Unable to extract column name when DIV exists


chezou:
While this is an upstream issue of sqlparse, `DIV` isn't treated as Operator, so it breaks column level parsing.

https://github.com/andialbrecht/sqlparse/issues/663

## Actual
```sh
$ .venv/bin/sqllineage -l column -e "INSERT OVERWRITE TABLE tab2
SELECT cast(conv(col1, 16, 10) as bigint) div 5 AS col2
FROM tab1"
<default>.tab2.div <- <default>.tab1.col1
```

## Expected

```sh
$ .venv/bin/sqllineage -l column -e "INSERT OVERWRITE TABLE tab2
SELECT cast(conv(col1, 16, 10) as bigint) div 5 AS col2
FROM tab1"
<default>.tab2.col2 <- <default>.tab1.col1
```

reata:
Using DIV as a arithmetic operator is not a universally supported feature among different SQL dialect and DIV is not a keyword in ANSI-2016. **Overall I think it's nice that we can support this feature but I don't want to do it by introducing DIV as keyword with monkey patching.** 

Referring to issues like #93 and #183 , adding DIV as keyword will deprive us the ability to use DIV as table name or column name without quoting, which is a valid use case for most of the dialects except for MySQL because DIV in MySQL is reserved keyword.

```sql
INSERT INTO TABLE div
SELECT col AS div
FROM foo;
```

Reference:

SQL Dialect  | support DIV operator | keyword | Reference
-- | -- | -- | --
ANSI-2016 | N | not a keyword | https://en.wikipedia.org/wiki/SQL_reserved_words
MySQL | Y | reserved | https://dev.mysql.com/doc/refman/8.0/en/keywords.html
PostgreSQL | N | not a keyword | https://www.postgresql.org/docs/current/sql-keywords-appendix.html
Oracle | N | not a keyword | https://docs.oracle.com/cd/A97630_01/appdev.920/a42525/apb.htm
Hive | Y | not a keyword* | https://cwiki.apache.org/confluence/display/hive/languagemanual+ddl https://cwiki.apache.org/confluence/display/hive/languagemanual+udf
SparkSQL | Y | Non-reserved | https://spark.apache.org/docs/latest/sql-ref-ansi-compliance.html
Trino | N | not a keyword | https://trino.io/docs/current/language/reserved.html

**support DIV operator**: Is SQL `SELECT 7 DIV 5` valid syntax?
[...]

Issue #233, opened 2022-03-07
Using JOIN with ON/USING keyword fails to determine source tables when followed by a parenthesis


snowmanwangyal:
I have the following queries:

```
create table target_table as (
    select *
    from source_table_a
    join source_table_b
    on (source_table_a.some_id = source_table_b.some_id))
)
```
```
create table target_table as (
    select *
    from source_table_a
    join source_table_b
    using (some_id)
)
```
This causes the table lineage to not extract `source_table_a` and `source_table_b` as sources. However, when the parenthesis are removed around `source_table_a.some_id = source_table_b.some_id)` and `some_id` it works properly.

Any help or insight would be greatly appreciated! 😀

snowmanwangyal:
I didn't spend too much time investigating, but I noticed from the `sqlparse` side when it is parsing the queries, it is treating everything inside as a Parenthesis whereas in the queries without the parenthesis it is treated as an Identifier.

I wonder where the logic lies where not having an identifier or having a parenthesis breaks it.

reata:
> I didn't spend too much time investigating, but I noticed from the `sqlparse` side when it is parsing the queries, it is treating everything inside as a Parenthesis whereas in the queries without the parenthesis it is treated as an Identifier.
> 
> I wonder where the logic lies where not having an identifier or having a parenthesis breaks it.

Generally it's OK for sqlparse to generate AST like that because it's a non-validating parser. We already have a bunch of logic in sqllineage to handle various edge cases like this. 

This issue is potentially related with #213

reata:
Confirmed this issue is introduced by PR #223 when I tried to strip the nested parenthesis around subquery. It's released with v1.3.4. Version 1.3.3 is fine. 


More details:
```sql
SELECT * FROM ((SELECT * FROM foo))
```

We have a buggy implementation by not asserting the parenthesis to be stripped at the beginning of the token. So in this case:
```sql
create table target_table as (
    select *
    from source_table_a
    join source_table_b
    on (source_table_a.some_id = source_table_b.some_id))
)
```

when inner parenthesis present after JOIN (or USING), it's identified as parenthesis around subquery, which is clearly wrong.

chezou:
Sorry for jumping in @reata, but do you have any plan to release this fix to PyPI?
I hit the same issue with 1.3.4 and it'd be great if I could have the fixed version wheel via PyPI.

reata:
[...]

Issue #183, opened 2021-11-13
Exception for Non-Reserved Keyword Used as Column Name 


reata:
```sql
INSERT OVERWRITE TABLE tab1
SELECT CASE WHEN locate('.', a.col1) > 0 THEN substring_index(a.col1, '.', 3) END host
FROM tab2 a
```

Non-reserved keyword: host, desc, summary for example.

reata:
Note TPS-DS query49.sql also suffers from this issue when non-reserve keyword catalog used as CTE name.

reata:
Let's fix #93 first, because in essence these two issues are the same thing.

reata:
It's working now, as we can use sqlfluff sparksql dialect for parsing.

PR #259, opened 2022-05-08
fix: Treat DIV as Operator

chezou:
This is a monkey patch covers #257 

Upstream PR is: https://github.com/andialbrecht/sqlparse/pull/664

### DIFF
--- a/sqllineage/__init__.py
+++ b/sqllineage/__init__.py
@@ -16,10 +16,18 @@ def _patch_adding_builtin_type() -> None:
     KEYWORDS["DATETIME"] = tokens.Name.Builtin
 
 
+def _patch_adding_operator_type() -> None:
+    from sqlparse import tokens
+    from sqlparse.keywords import KEYWORDS
+
+    KEYWORDS["DIV"] = tokens.Operator
+
+
 def _monkey_patch() -> None:
     try:
         _patch_adding_window_function_token()
         _patch_adding_builtin_type()
+        _patch_adding_operator_type()
     except ImportError:
         # when imported by setup.py for constant variables, dependency is not ready yet
         pass
--- a/tests/test_columns.py
+++ b/tests/test_columns.py
@@ -838,3 +838,18 @@ FROM tab1"""
             ),
         ],
     )
+
+
+def test_column_with_div():
+    sql = """INSERT OVERWRITE TABLE tab2
+SELECT cast(conv(col1, 16, 10) as bigint) div 5 AS col2
+FROM tab1"""
+    assert_column_lineage_equal(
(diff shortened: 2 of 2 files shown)


reata (commented):
DIV is keyword only in MySQL, right?

In that case, I think we should be very cautious because this potentially removes the possibility of parsing DIV as a table/column name.

Any thoughts?

chezou:
Unfortunately no. Hive and other related SQL also support it.

https://cwiki.apache.org/confluence/plugins/servlet/mobile?contentId=27362046#content/view/27362046
https://docs.databricks.com/sql/language-manual/functions/div.html

In my use case, I want use it for Hive.

reata:
Well, typically we support ANSI standard keyword. But Hive/Spark/Presto can be a exception since a lot of people are using these stacks.

As you mentioned, moving to KEYWORDS_HQL won't solve the problem. `sqlparse` doesn't provide a way to distinguish between reserved and non-reserved keywords, or keyword from specific SQL dialect. This is something we have to live with. At the same time, I have to worry about issues like #93 and #183 .

Allow me some more time to think. Will get back to you in this weekend.

BTW, I just released v1.3.5 because many user are stuck with the other parenthesis issue #233 . But don't worry for this PR, we can surely do more frequent releases at least for v1.3.6 coming after.

chezou:
Gotcha. I understand your concerns.

There can be several options to tackle with:
1. Enforce to quote table names or column names regardless of SQL dialect. This may be the easiest option for contributors, but sometime user may struggle with unrelated keywords in MySQL or Hive, etc.
2. Add `dialect` option like `--dialect mysql` and manage `KEYWORDS_HQL` properly. And tweak `sqlparse.keywords.is_keyword()` function with a monkey patch, like `is_keyword(dialect="hive")`. This can be flexible for users, but pain for maintainers.

I personally prefer option 1, because ignorance of the operator makes an entire parse breaking. But, it's up to your decision.

reata:
Sorry I came back to this a month later instead of a week. The project I'm driving now at work is a total nightmare and I've been working overtime for that.

I left a detailed comment in #257 and I think we can continue discussing there. In short, adding it as keyword might break other things and I don't think this feature is universal enough for me to do that. I was thinking maybe we can monkey patch the group_operator function to identify the T_NUMERICAL + DIV + T_NUMERICAL pattern, but I don't know if that will work.

For the SQL dialect and parser thing, my idea is that to some point we might need to replace sqlparse. There're already certain issues that's difficult, if not impossible, to solve with current parser. That's in the long term. I might come to do a PoC with ANTLR4 later this year.

PR #279, opened 2022-08-04
Update holders.py

johnclyde:
I'd like to be able to subclass and override SQLLineageHolder.of() in my code, but without returning the DiGraph object my entire method is copy-paste.

### DIFF
--- a/sqllineage/core/holders.py
+++ b/sqllineage/core/holders.py
@@ -222,7 +222,7 @@ class SQLLineageHolder(ColumnLineageMixin):
         }
 
     @staticmethod
-    def of(*args: StatementLineageHolder):
+    def of_digraph(*args: StatementLineageHolder) -> DiGraph:
         """
         To assemble multiple :class:`sqllineage.holders.StatementLineageHolder` into
         :class:`sqllineage.holders.SQLLineageHolder`
@@ -276,4 +276,13 @@ class SQLLineageHolder(ColumnLineageMixin):
         for node in [n for n, deg in g.degree if deg == 0]:
             if isinstance(node, Column) and len(node.parent_candidates) > 1:
                 g.remove_node(node)
+        return g
+
+    @staticmethod
+    def of(*args: StatementLineageHolder):
+        """
+        To assemble multiple :class:`sqllineage.holders.StatementLineageHolder` into
+        :class:`sqllineage.holders.SQLLineageHolder`
+        """
+        g = SQLLineageHolder.of_digraph(*args)
         return SQLLineageHolder(g)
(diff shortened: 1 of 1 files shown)


reata:
Sorry for the late reply, I was out for medical reasons. And thanks for your contribution.

I'm actually very interested in how you're going to leverage the subclass and override. Can you share more details?

Also for me DiGraph is something I'd like to hide. If we're going to merge this PR, please change **of_digraph** to a private method, something like **_build_digraph** makes more sense to me.

johnclyde:
> Sorry for the late reply, I was out for medical reasons. And thanks for your contribution.
> 
> I'm actually very interested in how you're going to leverage the subclass and override. Can you share more details?
> 
> Also for me DiGraph is something I'd like to hide. If we're going to merge this PR, please change **of_digraph** to a private method, something like **_build_digraph** makes more sense to me.

I'm happy to rename it to _build_digraph. We have had a fork of your library for several months to support database.schema.table_name, although I was recently told that you support this by pre-pending database to the schema.

It's becoming inconvenient to maintain a fork so I was trying to subclass your logic and in this case, subclass the SQLLineageHolder, but I had to copy/paste this whole method for now.

As an aside, I think you have a library and an app bundled into 1 package. The library doesn't need flask but it's coming along with the ride if we install your package with the app. Any thoughts about decoupling this?

reata:
> > Sorry for the late reply, I was out for medical reasons. And thanks for your contribution.
> > I'm actually very interested in how you're going to leverage the subclass and override. Can you share more details?
> > Also for me DiGraph is something I'd like to hide. If we're going to merge this PR, please change **of_digraph** to a private method, something like **_build_digraph** makes more sense to me.
> 
> I'm happy to rename it to _build_digraph. We have had a fork of your library for several months to support database.schema.table_name, although I was recently told that you support this by pre-pending database to the schema.
> 
> It's becoming inconvenient to maintain a fork so I was trying to subclass your logic and in this case, subclass the SQLLineageHolder, but I had to copy/paste this whole method for now.
> 
> As an aside, I think you have a library and an app bundled into 1 package. The library doesn't need flask but it's coming along with the ride if we install your package with the app. Any thoughts about decoupling this?

Thank you for the details and this helps. 

As for the mix of a library and application, indeed I was asked before. For me, flask is not a thing. We can make it as extras_require in setup.py any time we want, and asking people to do something like `pip install sqllineage[webapp]` only if they want an application bundle, otherwise it will be a pure python library.

The real headache is with the frontend bundle. There's no way I'm aware of to make it optional. If we can't do it, then the frontend bundle will eat up disk space anyway. And I would hate it if we separate the two parts into two repos/packages. 

Let me know if you have some good ideas to deal with it. Feel free to open an issue for discussion.

PR #281, opened 2022-08-11
add lineage expression to column lineages

IterableTrucks:
To get more precise column lineages, the Column model can be added with an expression field.

### DIFF
--- a/sqllineage/core/models.py
+++ b/sqllineage/core/models.py
@@ -18,7 +18,7 @@ from sqlparse.sql import (
 from sqlparse.utils import imt
 
 from sqllineage.exceptions import SQLLineageException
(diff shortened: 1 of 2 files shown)


reata:
@IterableTrucks Sorry for the late reply, I was out for medical reasons. And thanks for contributing.

I just approved the GitHub Action run, please fix the CI failure first. Also can you open an issue to briefly describe what you want to achieve here? That will be helpful.

reata (commented):
The overall code design and structure looks good to me. Since we're not explicitly exposing the expression attribute as public, it's acceptable without test cases written. But I'd like to try a few  cases to see what the expression result would be like. If everything goes smooth, I'll come back to merge this later this week. Thank you for your patience.

reata (approved):
LGTM

Issue #290, opened 2022-09-05
SQLLineageException when sourcing view from OPENROWSET() (T-SQL)


dvasdekis:
As mentioned in my other ticket, this library rules. Thank you again :)

T-SQL (as used in Azure Synapse Serverless) lets us define a view as being a file that sits on a filesystem. An example of such a view is the below:
```sql 
CREATE OR ALTER VIEW [etdl_raw].[exo_calibration_metadata] AS 
    SELECT  *  FROM
    -- Your parser bugs out here:
        OPENROWSET(
            BULK (
                    '/lake-raw/exo_calibration/metadata/**'
                    , '/lake-raw/this_file_doesnt_exist.parquet'
                )
            ,DATA_SOURCE = 'datasource_etdllake'
            ,FORMAT = 'Parquet'
    -- Contents of the file columns are defined in the below:
    ) WITH (
        [filepath] VARCHAR(255)
        ,[last_calibration_time] VARCHAR(255)
        ,[calibration_start_time] VARCHAR(255)
        ,[calibration_end_time] VARCHAR(255)
    ) AS t
    ;
GO
```

Currently, this gives us the below output:

```bash
PS C:\proj> sqllineage -f '.\above_file.sql'
Traceback (most recent call last):
  File "C:\Python310\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Python310\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "C:\Python310\Scripts\sqllineage.exe\__main__.py", line 7, in <module>
  File "C:\Python310\lib\site-packages\sqllineage\cli.py", line 86, in main
    runner.print_table_lineage()
  File "C:\Python310\lib\site-packages\sqllineage\runner.py", line 166, in print_table_lineage
    print(str(self))
  File "C:\Python310\lib\site-packages\sqllineage\runner.py", line 22, in wrapper
    self._eval()
  File "C:\Python310\lib\site-packages\sqllineage\runner.py", line 179, in _eval
    self._stmt_holders = [LineageAnalyzer().analyze(stmt) for stmt in self._stmt]
  File "C:\Python310\lib\site-packages\sqllineage\runner.py", line 179, in <listcomp>
    self._stmt_holders = [LineageAnalyzer().analyze(stmt) for stmt in self._stmt]
  File "C:\Python310\lib\site-packages\sqllineage\core\analyzer.py", line 60, in analyze
[...]

PR #291, opened 2022-09-05
Add fix for #290

dvasdekis:
Adds a simple pass to skip handing of OPENROWSET statements within the view definition.

### DIFF
--- a/sqllineage/core/handlers/source.py
+++ b/sqllineage/core/handlers/source.py
@@ -74,6 +74,12 @@ class SourceHandler(NextTokenBaseHandler):
                 self._handle(token.tokens[1], holder)
         elif token.ttype == Literal.String.Single:
             self.tables.append(Path(token.value))
+        elif (
+            type(token).__name__ == "Function"
+            and str(token)[0:10].lower() == "openrowset"
+        ):
+            # Handle OPENROWSET in T-SQL (Issue #290) - allows views to SELECT from filesystems on T-SQL
+            pass
         else:
             raise SQLLineageException(
                 "An Identifier is expected, got %s[value: %s] instead."
--- a/tests/test_others.py
+++ b/tests/test_others.py
@@ -292,3 +292,26 @@ def test_split_statements_with_desc():
 
 DESC tab1;"""
     assert len(LineageRunner(sql).statements()) == 2
+
+
+def test_create_table_with_openrowset():
+    # Check for issue #290 (https://github.com/reata/sqllineage/issues/290)
+    assert_table_lineage_equal(
+        """
+        CREATE OR ALTER VIEW tbl1 AS
+        SELECT  *  FROM
+            OPENROWSET(
+                BULK ( '/container/thing/**'
+                     , '/lake-raw/this_file_doesnt_exist.parquet' )
+                ,DATA_SOURCE = 'some_datasource', FORMAT = 'Parquet'
+        -- Contents of the file columns are defined in the below:
+        ) WITH (
+            [filepath] VARCHAR(255)
+            ,[last_calibration_time] VARCHAR(255)
+            ,[calibration_start_time] VARCHAR(255)
+            ,[calibration_end_time] VARCHAR(255)
+        ) AS t
+        ;""",
+        None,
+        {"tbl1"},
+    )


sqllineage/core/handlers/source.py
```diff
@@ -74,6 +74,12 @@ def _handle_table(self, token: Token, holder: SubQueryLineageHolder) -> None:
                 self._handle(token.tokens[1], holder)
         elif token.ttype == Literal.String.Single:
             self.tables.append(Path(token.value))
+        elif (
+            type(token).__name__ == "Function"
+            and str(token)[0:10].lower() == "openrowset"
+        ):
+            # Handle OPENROWSET in T-SQL (Issue #290) - allows views to SELECT from filesystems on T-SQL
+            pass
```
reata, line 82: Please rewrite it in a pythonic way , `isinstance(token, Function) and token.normalized.startswith("OPENROWSET")` makes sense to me. 

Also as mentioned in the other comment. Please get the path and make it as source. You can leverage get_parameters method in Function class.

reata (changes requested):
If we're going to support `SELECT  *  FROM OPENROWSET`, please get the path as source.

reata:
Please fix the conflicts first. This branch cannot be merge to master for the moment.

.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+    "python.testing.pytestArgs": [
+        "tests"
+    ],
+    "python.testing.unittestEnabled": false,
+    "python.testing.pytestEnabled": true
+}
```
reata, line 7: Please exclude this file as we also exclude PyCharm configs.

sqllineage/core/handlers/source.py
```diff
@@ -55,8 +55,65 @@ def _handle(self, token: Token, holder: SubQueryLineageHolder) -> None:
...
+        Holds logic to handle openrowset token.
+        """
+        format_statement_found = False
+        datasource_statement_found = False
+        datasource_string_value = ""
+        tokens_to_append_to_paths = []
+        for (
+            each
+        ) in (
+            token.flatten()
```
reata, line 69: Any possibility to traverse the tree instead of flatten?

sqllineage/utils/helpers.py
```diff
@@ -5,6 +5,9 @@
 
 
 def escape_identifier_name(name: str):
+    """
+    Removes singlequotes ('), doublequotes (") and backticks (`) from a string. Used to sanitise file paths.
+    """
```
reata, line 10: It's not used only for paths. Mainly for table names. The comment here is inaccurate.

reata (changes requested):
Please fix the conflict and also addressing the inline comment.

dvasdekis:
> Please fix the conflicts first. This branch cannot be merge to master for the moment.
It's saying my branch is up to date currently? Also no conflicts are reporting in this PR for me.

reata (changes requested):
Sorry for misleading you. I should be able to squash and merge. (conflict is with rebase and merge). Anyway, don't worry about the conflict.

But please fix the issues I mentioned in inline comment. Especially on token.flatten() part. We should try take advantage of the tree structure, use Function class, instead of flatten all the inner tokens, unless we have to.

dvasdekis:
> Sorry for misleading you. I should be able to squash and merge. (conflict is with rebase and merge). Anyway, don't worry about the conflict.
> 
> But please fix the issues I mentioned in inline comment. Especially on token.flatten() part. We should try take advantage of the tree structure, use Function class, instead of flatten all the inner tokens, unless we have to.

No problem. Do you know an example of where tree traversal has been used with sqlparse library? The docs are pretty bad!

reata:
> > Sorry for misleading you. I should be able to squash and merge. (conflict is with rebase and merge). Anyway, don't worry about the conflict.
> > But please fix the issues I mentioned in inline comment. Especially on token.flatten() part. We should try take advantage of the tree structure, use Function class, instead of flatten all the inner tokens, unless we have to.
> 
> No problem. Do you know an example of where tree traversal has been used with sqlparse library? The docs are pretty bad!

I feel the same about the docs. Basically I was learning by reading the code.

One other thing I'd like to mention is that I'm working on replace sqlparse with antlr generated parser. The reason is documented in #302 . So don't push yourself too hard if you face difficulty dealing with sqlparse.

reata:
Closing this for now, as we need a sqlfluff implementation as well as sqlparse with v1.4.x released.

PR #293, opened 2022-09-05
Fix #292 with pass statement

### LINKED ISSUE #292: SQLLineageException when CREATE VIEW defined as SELECT ... FROM ( VALUES ... )
Similar to #290 , when a CREATE VIEW statement is defined as either example present in the below:

```sql
CREATE OR ALTER VIEW tbl1 AS
    SELECT  cast(my_overrides.col1 as varchar(100)) as col1
   , cast(my_overrides.col2 as varchar(100)) as col2
   , cast(my_overrides.col3 as datetime2) as col3
   , cast(my_overrides.col4 as datetime2) as col4
   , cast(my_overrides.col5 as smallint) as col5
   , cast(my_overrides.col6 as varchar(1024)) as col6
    FROM  (VALUES
     ('data','thing1','1971-01-01 00:00:00', '1971-01-01 00:00:10', '91', 'some text')
   , ('data','thing2','1971-01-01 00:00:10', '1971-01-01 00:00:20', '92', 'some text')
   , ('data','thing3','1971-01-01 00:00:10', '1971-01-01 00:00:20', '92', 'some text')   
 )my_overrides(col1, col2, col3, col4, col5, col6) ;

-- same result but with inferred column names and types
CREATE OR ALTER VIEW tbl2 AS
    SELECT * FROM  (VALUES
     ('data','thing1','1971-01-01 00:00:00', '1971-01-01 00:00:10', '91', 'some text')
   , ('data','thing2','1971-01-01 00:00:10', '1971-01-01 00:00:20', '92', 'some text')
   , ('data','thing3','1971-01-01 00:00:10', '1971-01-01 00:00:20', '92', 'some text')
 )my_overrides(col1, col2, col3, col4, col5, col6) ;
```

The result is 
```bash
PS C:\proj> sqllineage -f '.\try.sql'
Traceback (most recent call last):
  File "C:\Python310\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Python310\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "C:\Python310\Scripts\sqllineage.exe\__main__.py", line 7, in <module>
  File "C:\Python310\lib\site-packages\sqllineage\cli.py", line 86, in main
    runner.print_table_lineage()
  File "C:\Python310\lib\site-packages\sqllineage\runner.py", line 166, in print_table_lineage
    print(str(self))
  File "C:\Python310\lib\site-packages\sqllineage\runner.py", line 22, in wrapper
    self._eval()
  File "C:\Python310\lib\site-packages\sqllineage\runner.py", line 179, in _eval
    self._stmt_holders = [LineageAnalyzer().analyze(stmt) for stmt in self._stmt]
  File "C:\Python310\lib\site-packages\sqllineage\runner.py", line 179, in <listcomp>
    self._stmt_holders = [LineageAnalyzer().analyze(stmt) for stmt in self._stmt]
  File "C:\Python310\lib\site-packages\sqllineage\core\analyzer.py", line 60, in analyze
    self._extract_from_dml(stmt, AnalyzerContext())
  File "C:\Python310\lib\site-packages\sqllineage\core\analyzer.py", line 135, in _extract_from_dml
    next_handler.handle(sub_token, holder)
  File "C:\Python310\lib\site-packages\sqllineage\core\handlers\base.py", line 39, in handle
    self._handle(token, holder)
  File "C:\Python310\lib\site-packages\sqllineage\core\handlers\source.py", line 56, in _handle
    self._handle_table(token, holder)
  File "C:\Python310\lib\site-packages\sqllineage\core\handlers\source.py", line 74, in _handle_table
    self._handle(token.tokens[1], holder)
  File "C:\Python310\lib\site-packages\sqllineage\core\handlers\source.py", line 56, in _handle
    self._handle_table(token, holder)
  File "C:\Python310\lib\site-packages\sqllineage\core\handlers\source.py", line 78, in _handle_table
    raise SQLLineageException(
sqllineage.exceptions.SQLLineageException: An Identifier is expected, got Token[value: VALUES] instead.
```

I hope it's OK to treat it in a similar way to #291 as this is what seems to be the right approach (just ignore it)? Is there any information of value that you would like to scrap from the VALUES statement? (e.g. column-based lineage?). I'll assume not for now, but please let me know if so.

Thanks again!

dvasdekis:
Fixes #292 by adding a pass statement when the Token in _handle_table is "values", implying a `CREATE TABLE AS SELECT * FROM VALUES (...)` statement.


sqllineage/core/handlers/source.py
```diff
@@ -74,6 +74,9 @@ def _handle_table(self, token: Token, holder: SubQueryLineageHolder) -> None:
                 self._handle(token.tokens[1], holder)
         elif token.ttype == Literal.String.Single:
             self.tables.append(Path(token.value))
+        # If we have a CREATE TABLE AS SELECT * FROM VALUES (...), this stops the parser erroring out:
+        elif type(token).__name__ == "Token" and str(token)[0:6].lower() == "values":
+            pass
```
reata, line 79: The token here is a instance of Parenthesis. Please reuse that elif above. Also, make the function call pythonic. Don't use __name__ or str casting when we have alternative choice.

reata (changes requested):

sqllineage/core/handlers/source.py
dvasdekis: Thank you for your great feedback, it really helps me to learn too!

I have tried to use the Parenthesis IF [from here](https://github.com/reata/sqllineage/blob/e67293a874e34f5d11baa88aa47b3aa7af99f1e4/sqllineage/core/handlers/source.py#L58), but I get failed tests if I do this:
```python
        elif isinstance(token, Parenthesis):
            if is_subquery(token):
                # SELECT col1 FROM (SELECT col2 FROM tab1), the subquery will be parsed as Parenthesis
                # This syntax without alias for subquery is invalid in MySQL, while valid for SparkSQL
                self.tables.append(SubQuery.of(token, None))
            elif str(token)[0:6].lower() == "values":
                # If we have a CREATE TABLE AS SELECT * FROM VALUES (...), this stops the parser erroring out:
                pass
            else:
                # SELECT * FROM (tab2), which is valid syntax
                self._handle(token.tokens[1], holder)
        elif token.ttype == Literal.String.Single:
            self.tables.append(Path(token.value))
        else:
            raise SQLLineageException(
                "An Identifier is expected, got %s[value: %s] instead."
                % (type(token).__name__, token)
            )
```

failed test: 
```
FAILED tests/test_others.py::test_view_as_select_from_values_with_colnames - sqllineage.exceptions.SQLLineageException: An Identifier is expected, got Token[value: VALUES] instead.
FAILED tests/test_others.py::test_view_as_select_from_values_no_colnames - sqllineage.exceptions.SQLLineageException: An Identifier is expected, got Token[value: VALUES] instead.
```

Sorry if I'm missing something obvious, I haven't seen these custom subtypes used before and I don't yet know how to run SQL against your library and see which one an input gets detected as.

reata: Yes this is the right place. You can look at how we decide a Parenthesis is subquery or not. Basically it's finding the first token and see whether it's SELECT. Same here, we'd better fill in a logic to make sure the first token within Parenthesis is keyword VALUES, only after that can be sure that this can be skipped. 

As for 'how to run SQL against your library and see which one an input gets detected as', you can use an debugger or simply add a print statement. The token type and token class come from sqlparse library. You can also refer to its document.

dvasdekis:
Hi, it still isn't coming through as a parenthesis (only as a token), and I don't think there's any case where the parenthesis is an issue, so I'm hoping this code is OK. Thank you.

dvasdekis:
I switched to a Pythonic token test as per the comment in your other review. Thanks for the help!

sqllineage/core/handlers/source.py
dvasdekis: I switched to a Pythonic token test as per the comment in your other review. Thanks for the help!

sqllineage/core/handlers/source.py
```diff
@@ -72,6 +72,9 @@ def _handle_table(self, token: Token, holder: SubQueryLineageHolder) -> None:
             else:
                 # SELECT * FROM (tab2), which is valid syntax
                 self._handle(token.tokens[1], holder)
```
reata, line 74: The token is of type Parenthesis. So it goes here first. But in our case, it's not like what's mentioned in the comment. In our case, the value of this token is
```
(VALUES\n     ('data','thing1','1971-01-01 00:00:00', '1971-01-01 00:00:10', '91', 'some text')\n   , ('data','thing2','1971-01-01 00:00:10', '1971-01-01 00:00:20', '92', 'some text')\n   , ('data','thing3','1971-01-01 00:00:10', '1971-01-01 00:00:20', '92', 'some text')\n )
```
It's meaningless to get the index=1 sub_token. We really should add a elif in line #72 to decide whether the index=1 token is VALUE keyword or not.

sqllineage/core/handlers/source.py
```diff
@@ -72,6 +72,9 @@ def _handle_table(self, token: Token, holder: SubQueryLineageHolder) -> None:
...
+        elif token.normalized.startswith("VALUES"):
```
reata, line 75: Once we get the token as VALUE, add another safeguard to make sure token.is_keyword is True. Also, let's make the existing expression token.normalized == "VALUES"

reata (changes requested):
Sorry for the late reply. It's been a busy week.
The control flow here is not as expected, although the test case passes. Please refer to the inline comment.

sqllineage/core/handlers/source.py
dvasdekis: Thank you! That works great. :)

reata (approved):
LGTM. Thanks for your effort and congrats on your first PR merged.

Issue #302, opened 2022-10-12
Replace sqlparse With sqlfluff


reata:
As of v1.3.6, I think table level lineage is perfectly production-ready. For column level lineage, under the no-metadata background, it's also as good as it can be. And yet we still have a lot of corner cases that are not yet supported. This is really due to the longtail of SQL language features and fragmentation of various SQL dialect.

Typical issues:

- dialect specific function: 
  - #301  
  - #272 
  - #248  
  - #214 
- dialect specific syntax:  
  - #205 
  - #166
- nested parenthesis: 
  - #278 
  - #249 
  - #244 
- identifier escape: 
  - #218 
- reserve/non-reserve keyword: 
  - #183
  -  #93 
- syntax validation: 
  - #159 

It's almost impossible to solve all of the issues. Some of them contradict with each other. 

Over the years, we already have several monkey patches and utils on sqlparse (see https://github.com/reata/sqllineage/blob/master/sqllineage/__init__.py and https://github.com/reata/sqllineage/blob/master/sqllineage/utils/sqlparse.py). Due to the non-validating nature of sqlparse, that's the bitter pill to swallow when we enjoyed tons of convenience.

But to move forward, I believe we come to a point that we need more control of underlying parser so that:

1) we know better what syntax/language feature we support.
2) we can easily revise the AST to our need when we decide to support some new features.

To kick off the process, I have already write a POC project with antlr4 at https://github.com/reata/sqllineage4j. All the table level lineage is ready there and it takes much less effort than I expected. Using antlr4 means we can refer to the source code from Spark or Presto/Trino. After I verify that column lineage can also be analyzed with ease, I'll port that version back here. This will be a step for v1.4.0 version.

reata:
I'd like to mention that community is also trying the parser underneath sqlfluff. The antlr4 is not the final decision yet. We'll evaluate based on the progress.

reata:
With #326 merged into master, we now have a new `dialect` option. When passed with real dialect, like mysql, oracle, hive, sparksql, bigquery, snowflake, etc, we'll leverage sqlfluff to analyze the query. A pseudo dialect `non-validating` is introduced to remain backward compatibility, falling back to use sqlparse as parser. 

We have dual test using both parser and make sure the lineage result is exactly the same for every test case except for the following three:
| test case | dialect |
| ---- | ---- |
| test_create_using_serde | hive |
| test_copy_into_path | tsql |
| test_select_from_unnest_with_ordinality | athena |

[...]

PR #326, opened 2023-01-13
Add an option to use `sqlfluff` to be used as a query parser

nahuelverdugo:
Hi @reata,

In this PR, we add a new flow where the user can start using a different parser, `sqlfluff`, for computing the lineage. The main advantage of using `sqlfluff` in the analyzer is to get better query parsing out of the box without having to rewrite multiple SQL specifications. The idea of this first PR is to replicate the same behavior that SQLLineage was using with `sqlparse`.

To achieve this, we have added the following:
- Extra parameters in the LineageRunner for accepting a dialect as a parameter and another one for using the new flow with `sqlfluff`
- Multiple query extractors depending on the query type, e.g., DML select, DDL drop, CTE, etc
- Added a model and holder specific for `sqlfluff`
- A set of utils to deal with `sqlfluff` objects
- Tests to test new and old flows and compare the graphs generated by both implementations
 
In addition to these changes, we also have done the following:
- Removed the compatibility with Python 3.7 since `sqlfluff` does not support version 3.6 
- Updated flake8 to the Python 3.7 version
- ~Changed port 5000 to 5001 for the web application. It can not be used in MacOS when Airplay Receiver is enabled.~

We think this could be a good starting point to test how `sqlfluff` behaves with SQLLineage and an opportunity to contribute to both projects :)


sqllineage/__init__.py
```diff
@@ -74,4 +74,6 @@ def _monkey_patch() -> None:
     "SQLLINEAGE_DIRECTORY", os.path.join(os.path.dirname(__file__), "data")
 )
 DEFAULT_HOST = "localhost"
-DEFAULT_PORT = 5000
+DEFAULT_PORT = 5001
```
reata, line 77: We already have an option --port to let user specify PORT used for web UI, let's not change the default port here.

sqllineage/drawing.py
```diff
@@ -152,7 +158,9 @@ def lineage(payload):
 
     req_args = Namespace(**payload)
     sql = extract_sql_from_args(req_args)
-    lr = LineageRunner(sql, verbose=True)
+    dialect = getattr(req_args, "dialect", DEFAULT_DIALECT)
+    use_sql_parse = bool(getattr(req_args, "use_sqlfluff", DEFAULT_USE_SQLFLUFF))
+    lr = LineageRunner(sql, verbose=True, dialect=dialect, use_sqlparse=use_sql_parse)
```
reata, line 163: For Web UI lineage visualization, it's all forced to use SQLFLUFF as parser, correct? As payload here is passed from frontend and I don't see any frontend code update.

Later if we decide to only keep sqlfluff, it could be a good idea to add dialect option in frontend UI.

tests/test_select.py
```diff
@@ -16,11 +16,13 @@ def test_select_with_schema_and_database():
...
 def test_select_with_table_name_in_backtick():
-    assert_table_lineage_equal("SELECT * FROM `tab1`", {"tab1"})
+    assert_table_lineage_equal("SELECT * FROM `tab1`", {"tab1"}, dialect="bigquery")
 
 
 def test_select_with_schema_in_backtick():
-    assert_table_lineage_equal("SELECT col1 FROM `schema1`.`tab1`", {"schema1.tab1"})
+    assert_table_lineage_equal(
+        "SELECT col1 FROM `schema1`.`tab1`", {"schema1.tab1"}, dialect="bigquery"
+    )
```
reata, line 25: We can use pytest parametrize functionality to test these two cases. backtick for quoting identifier is also a routine from MySQL as well as BigQuery.
```python
@pytest.mark.parametrize("dialect", ["mysql", "bigquery"])
def test_select_with_schema_in_backtick(dialect):
```

tests/test_select.py
```diff
@@ -147,16 +149,19 @@ def test_select_left_join_with_extra_space_in_middle():
     assert_table_lineage_equal("SELECT * FROM tab1 LEFT  JOIN tab2", {"tab1", "tab2"})
 
 
+# deactivated for sqlfluff since it can not be parsed properly
 def test_select_left_semi_join():
     assert_table_lineage_equal(
-        "SELECT * FROM tab1 LEFT SEMI JOIN tab2", {"tab1", "tab2"}
+        "SELECT * FROM tab1 LEFT SEMI JOIN tab2", {"tab1", "tab2"}, test_sqlfluff=False
```
reata, line 155: Looks like we can make this work by specifying dialect="hive", am I missing anything here?

tests/test_select.py
```diff
@@ -242,20 +250,23 @@ def test_select_from_unnest_parsed_as_keyword():
     )
 
 
+# deactivated for sqlfluff since it can not be parsed properly
 def test_select_from_unnest_with_ordinality():
```
reata, line 254: This will make presto/trino user unhappy. Let's revisit all the unsupported cases later together.

tests/test_insert.py
```diff
@@ -73,4 +86,21 @@ def test_insert_overwrite_from_self_with_join():
...
+as
+SELECT
+    col1,
+    col2
+FROM tab1
+GROUP BY
+col1""",
+        {"tab1"},
+        {"view1"},
+        "tsql",
```
reata, line 105: This is ANSI compatible, let's remove dialect

tests/test_insert.py
```diff
@@ -34,24 +36,34 @@ def test_insert_into_partitions():
...
     )
 
 
 def test_insert_overwrite_with_keyword_table():
     assert_table_lineage_equal(
-        "INSERT OVERWRITE TABLE tab1 SELECT col1 FROM tab2", {"tab2"}, {"tab1"}
+        "INSERT OVERWRITE TABLE tab1 SELECT col1 FROM tab2",
+        {"tab2"},
+        {"tab1"},
+        dialect="sparksql",
```
reata, line 57: I think the overall principle to test dialect specific feature is to list all dialect supported. In this case, we should test hive and sparksql.

Also maybe we should consider moving all the dialect specific test suits to a dedicated file.

tests/test_cte.py
```diff
@@ -19,6 +19,7 @@ def test_with_select_one_without_as():
     assert_table_lineage_equal(
         "WITH wtab1 (SELECT * FROM schema1.tab1) SELECT * FROM wtab1",
         {"schema1.tab1"},
+        dialect="mysql",
```
reata, line 22: I believe this is a bug from sqlfluff. This syntax is actually sparksql only. MySQL will throw syntax error.

tests/test_others.py
```diff
@@ -42,9 +46,10 @@ def test_create_as():
 
 
 def test_create_as_with_parenthesis_around_select_statement():
-    assert_table_lineage_equal(
-        "CREATE TABLE tab1 AS (SELECT * FROM tab2)", {"tab2"}, {"tab1"}
-    )
+    sql = "CREATE TABLE tab1 AS (SELECT * FROM tab2)"
+    assert_table_lineage_equal(sql, {"tab2"}, {"tab1"}, test_sqlfluff=False)
+    # Graph generated differ but it is correct
+    assert_table_lineage_equal(sql, {"tab2"}, {"tab1"}, test_sqlparse=False)
```
reata, line 52: duplication lines.

Also this test case works with hive and sparksql.

tests/test_others.py
```diff
@@ -54,9 +59,10 @@ def test_create_as_with_parenthesis_around_table_name():
 
 
 def test_create_as_with_parenthesis_around_both():
-    assert_table_lineage_equal(
-        "CREATE TABLE tab1 AS (SELECT * FROM (tab2))", {"tab2"}, {"tab1"}
-    )
+    sql = "CREATE TABLE tab1 AS (SELECT * FROM (tab2))"
+    assert_table_lineage_equal(sql, {"tab2"}, {"tab1"}, test_sqlfluff=False)
+    # Graph generated differ but it is correct
+    assert_table_lineage_equal(sql, {"tab2"}, {"tab1"}, test_sqlparse=False)
```
reata, line 65: duplication lines. Same as above.

tests/test_others.py
```diff
@@ -175,11 +196,13 @@ def test_rename_table():
     This syntax is MySQL specific:
      https://dev.mysql.com/doc/refman/8.0/en/rename-table.html
     """
-    assert_table_lineage_equal("rename table tab1 to tab2", None, None)
+    assert_table_lineage_equal("rename table tab1 to tab2", None, None, "teradata")
```
reata, line 199: See comment. This is also for mysql

tests/test_select.py
nahuelverdugo: Trino and Presto is still not supported by `sqlfluff` but opening a PR to support them should not be hard.

tests/test_insert.py
nahuelverdugo: I agree with this. I have added some `_dialect_specific.py` tests for those tests where we are not testing ANSI queries.

reata (commented):
testing module looks good to me now.

Shall we start removing sqlparse related code? I'd like to see what functionality left that we still have to rely on sqlparse. 

_Update: we decide to keep sqlparse while growing with sqlfluff. An interface/facade will be built to hide parser implementation details._

tests/test_cte.py
nahuelverdugo: I think for this case, we can open a PR in `sqlfluff`.

sqllineage/drawing.py
nahuelverdugo: I have tried to put some controls in the UI to select some dialects and use `sqlfluff` as the parser, but I don't know what is missing to make it work properly.

reata: This is lower priority for the moment. I can take a look later.

tests/test_select.py
reata: I just checked sqlfluff [dialect docs](https://docs.sqlfluff.com/en/stable/dialects.html), and it supports Athena. Athena is built on open-source Trino and Presto. So we can point Presto/Trino user to dialect="athena" for most cases.

Athena supports UNNEST: https://docs.aws.amazon.com/athena/latest/ug/flattening-arrays.html
But I don't see WITH CARDINALITY as mentioned in Presto docs supported in Athena: https://prestodb.io/docs/current/sql/select.html

If WITH CARDINALITY is really Presto only, then we have to walk-around. Otherwise, we can improve sqlfluff athena dialects.

Do you happen to have a Athena environment to try this SQL?

tests/test_cte.py
reata: For the moment, I have removed mysql from this test case despite it works. Only tested against databricks and sparksql because this is documented as a feature in their language manuals.

tests/test_columns.py
```diff
@@ -635,18 +659,16 @@ def test_cast_to_data_type(dtype):
     )
 
 
-@pytest.mark.parametrize(
-    "dtype", ["string", "timestamp", "date", "datetime", "decimal(18, 0)"]
-)
-def test_nested_cast_to_data_type(dtype):
-    sql = f"""INSERT OVERWRITE TABLE tab1
+@pytest.mark.parametrize("dtype", ["string", "timestamp", "date", "decimal(18, 0)"])
+def test_nested_cast_to_data_type(dtype: str):
```
reata, line 663: Can you take a look at this test case? It's extremely slow to run.

tests/test_columns.py
```diff
@@ -655,11 +677,9 @@ def test_nested_cast_to_data_type(dtype):
     )
 
 
-@pytest.mark.parametrize(
-    "dtype", ["string", "timestamp", "date", "datetime", "decimal(18, 0)"]
-)
-def test_cast_to_data_type_with_case_when(dtype):
-    sql = f"""INSERT OVERWRITE TABLE tab1
+@pytest.mark.parametrize("dtype", ["string", "timestamp", "date", "decimal(18, 0)"])
+def test_cast_to_data_type_with_case_when(dtype: str):
```
reata, line 681: Moderately slow.

tests/test_columns.py
```diff
@@ -883,8 +929,8 @@ def test_column_lineage_multiple_paths_for_same_column():
         "coalesce(col1, 0) as decimal(10, 6)",
     ],
 )
-def test_column_try_cast_with_func(func):
-    sql = f"""INSERT OVERWRITE TABLE tab2
+def test_column_try_cast_with_func(func: str):
```
reata, line 932: Moderately slow.

reata (commented):
Please take a look at performance issue regarding several column test cases to see if it's sqlfluff related or are we using it in a wrong way.

tests/test_columns.py
reata:
```python
def test_nested_cast_to_data_type():
    dtype = "string"
    sql = f"""INSERT INTO tab1
SELECT cast(cast(cast(cast(cast(col1 AS {dtype}) AS {dtype}) AS {dtype}) AS {dtype}) AS {dtype}) AS col1
FROM tab2"""
    assert_column_lineage_equal(
        sql,
        [(ColumnQualifierTuple("col1", "tab2"), ColumnQualifierTuple("col1", "tab1"))],
        test_sqlparse=False
    )
```

I tried this, and it's 1 sec 534 ms to run on my machine. Doesn't look like a dialect issue.

nahuelverdugo:
> The code structure is much better now. I was thinking about only keeping the shared parts (holders.py, models.py) in `sqllineage.core`, and move sqllineage.sqlfluff_core to `sqllineage.core.parser.sqlfluff`, meanwhile all the sqlparse specifc code to `sqllineage.core.parser.sqlparse`.
> 
> Do you think it's a good idea?

I was thinking about the same when doing the changes, so I prefer that code structure too. I have pushed a change to address it.

reata (commented):
1. sqllineage is actually about two things, 1) analyze the AST generated by a parser, 2) assemble the information into DAG leveraging a graph library. Ideally both parser and graph lib should be plug-gable. Given this context, **please still put sqlparse and sqlfluff modules under `sqllineage.core.parser`**. Later in separate PR we shall have `sqllineage.core.graph`. And to clarify, I have no intention to replace networkx, that is just to make the code more readable.

2. For sqllineage/core/sqlfluff/utils/entities.py, looks like we can completely remove it. See my comment inline.

3. Similar to the current sqllineage/core/sqlfluff/utils/sqlfluff.py, let's move sqllineage/utils/sqlparse.py under sqlparse folder since that is also parser specific. Keeping only constant.py, entities.py and helpers.py in sqllineage/utils as global utils.

4. Can we also try move monkey patches in sqllineage/__init_\_.py to sqlparse folder?

sqllineage/drawing.py
reata: Frontend code has been updated, together with draw.py. I merged the two settings (use_sqlfluff and dialect) into one unified dialect setting, by introducing a pseudo dialect called "non-validating" used for sqlparse only.

I'm still thinking about whether we should do the same on backend, by removing the use_sqlfluff option in LineageRunner. Let me know your thought.

reata:
@nahuelverdugo I have pushed changes to address dialect settings from frontend UI and overall code structure change regarding my latest comment. You can focus on bringing in UNION parsing functionality and the performance issue mentioned above next week. I think that's our top priority next phase.

nahuelverdugo:
@reata the UNION parsing functionality is done and I have reviewed the performance issue but not sure how we should proceed since the issue seems to be in the `sqlfluff` module when we have nested functions.

tests/test_select.py
nahuelverdugo: I have reviewed it on our Athena instance, and it works, so we can contribute to `sqlfluff` for fixing it and point Presto/Trino users to the `athena` dialect.

image

tests/test_columns.py
reata: OK, this is sqlfluff's issue. Won't be a blocker for this PR then. Let's resolve these. Maybe I'll try if I can submit a PR to them later.

reata:
Huge thanks for your effort. I'll make one final pass on this PR over the weekend. And let's target merge this into master next week.

tests/test_select.py
reata: Sorry I missed this thread. It's great news. I'll document this together with other things we'd want to improve sqlfluff as action items post this PR merge.

sqllineage/core/parser/sqlfluff/extractors/dml_select_extractor.py
```diff
@@ -0,0 +1,69 @@
...
+        # call end of query hook here as loop is over
+        for conditional_handler in conditional_handlers:
+            conditional_handler.end_of_query_cleanup(holder)
+
+        # By recursively extracting each subquery of the parent and merge, we're doing Depth-first search
+        for sq in subqueries:
+            holder |= self.extract(sq.query, AnalyzerContext(sq, holder.cte))
+
+        for sq in holder.extra_subqueries:
+            holder |= self.extract(sq.query, AnalyzerContext(sq, holder.cte))
```
reata, line 67: I don't quite understand the extra_subqueries part, could you explain a little bit?

reata:
An important change is made to remove use_sqlfluff option. As an alternative, a pseudo dialect "non-validating" is introduced (same as we did in frontend code) for compatibility. Whenever user explicitly pass this dialect, we'll fallback to sqlparse implementation. Otherwise we go with sqlfluff. This also makes it easy when later we deprecate sqlparse in v1.5.x release.

Other changes are purely structural to remove duplication and bring in more interface to hide parser details.

The core code is done. I need some more time to go through all the test cases once more to make sure we didn't miss anything.

nahuelverdugo:
Hi @reata, thank you for the update. IMHO, for the first iteration, we can leave the default behavior to fall into `sqlparse,`. Then, once we fix some dialect-specific issues and improve the `sqlfluff` flow, we can introduce the `non-validating` pseudo dialect.

Please let me know if something is missing, and I will address it as soon as possible.

reata:
Let's postpone make this decision on the default behavior. Just want to mention that merging this PR into master doesn't mean we're going for a v1.4.0 release immediately. I want to spend some more time evaluating the corner cases we used to not support as listed in #302 . The reasonable v1.4.0 release timeline will be in end of Q1. By that time we can discuss the strategy.

For this PR, I believe we're probably fine now. Just as a pre-caution I'd like to check the test cases. I got some heavy workload this week so it might still need to be done in weekend. Nothing you need to worry about.

reata:
Just merged this PR. Congratulations @nahuelverdugo @ulixius9 

I've listed all the next steps in #302 . If you have any update, please comment there. Again, thank you so much for making this happen.

nahuelverdugo:
Thank you, @reata! Let's keep collaborating and pushing to improve the library.

PR #360, opened 2023-04-18
Fix version of 'sqlparse' dependency to 0.4.4

nahuelverdugo:
With the latest `0.4.4` version of `sqlparse`, we started to see this error when doing the _monkey_patch() to the library:

```
env/lib/python3.9/site-packages/sqllineage/runner.py:9: in <module>
    from sqllineage.core.parser.sqlparse.analyzer import SqlParseLineageAnalyzer
env/lib/python3.9/site-packages/sqllineage/core/parser/sqlparse/__init__.py:34: in <module>
    _monkey_patch()
env/lib/python3.9/site-packages/sqllineage/core/parser/sqlparse/__init__.py:31: in _monkey_patch
    _patch_updating_lateral_view_lexeme()
env/lib/python3.9/site-packages/sqllineage/core/parser/sqlparse/__init__.py:21: in _patch_updating_lateral_view_lexeme
    if regex("LATERAL VIEW EXPLODE(col)"):
E   TypeError: 'str' object is not callable
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,7 +65,7 @@ setup(
     ],
     python_requires=">=3.8",
     install_requires=[
-        "sqlparse>=0.3.1",
+        "sqlparse==0.4.3",
         "networkx>=2.4",
         "sqlfluff>=1.4.5",
     ],
--- a/setup.py
+++ b/setup.py
@@ -65,9 +65,9 @@ setup(
     ],
     python_requires=">=3.8",
     install_requires=[
-        "sqlparse==0.4.4",
+        "sqlparse>=0.4.4",
         "networkx>=2.4",
-        "sqlfluff==2.0.3",
+        "sqlfluff>=2.0.0,<=2.0.3",
     ],
     entry_points={"console_scripts": ["sqllineage = sqllineage.cli:main"]},
     extras_require={
--- a/setup.py
+++ b/setup.py
@@ -65,9 +65,9 @@ setup(
     ],
     python_requires=">=3.8",
     install_requires=[
-        "sqlparse==0.4.3",
+        "sqlparse==0.4.4",
         "networkx>=2.4",
-        "sqlfluff>=1.4.5",
+        "sqlfluff==2.0.3",
     ],
     entry_points={"console_scripts": ["sqllineage = sqllineage.cli:main"]},
     extras_require={
--- a/sqllineage/core/parser/sqlparse/__init__.py
+++ b/sqllineage/core/parser/sqlparse/__init__.py
@@ -18,10 +18,10 @@ def _patch_adding_builtin_type() -> None:
 
 def _patch_updating_lateral_view_lexeme() -> None:
     for i, (regex, lexeme) in enumerate(SQL_REGEX):
-        if regex("LATERAL VIEW EXPLODE(col)"):
+        rgx = re.compile(regex, re.IGNORECASE | re.UNICODE).match
+        if rgx("LATERAL VIEW EXPLODE(col)"):
             new_regex = r"(LATERAL\s+VIEW\s+)(OUTER\s+)?(EXPLODE|INLINE|PARSE_URL_TUPLE|POSEXPLODE|STACK|JSON_TUPLE)\b"
-            new_compile = re.compile(new_regex, re.IGNORECASE | re.UNICODE).match
-            SQL_REGEX[i] = (new_compile, lexeme)
+            SQL_REGEX[i] = (new_regex, lexeme)
             break
 
 
--- a/sqllineage/core/parser/sqlparse/models.py
+++ b/sqllineage/core/parser/sqlparse/models.py
@@ -2,7 +2,6 @@ from typing import List, Optional
 
 from sqlparse import tokens as T
 from sqlparse.engine import grouping
-from sqlparse.keywords import is_keyword
 from sqlparse.sql import (
     Case,
     Comparison,
@@ -15,6 +14,7 @@ from sqlparse.sql import (
     TokenList,
 )
 from sqlparse.utils import imt
+from sqlparse.lexer import Lexer
 
 from sqllineage.core.models import Column, Schema, SubQuery, Table
 from sqllineage.core.parser.sqlparse.utils import get_parameters, is_subquery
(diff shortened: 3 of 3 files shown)


reata:
Looks like sqlparse just released v0.4.4 version and it breaks our monkey patching. 

Do you mind fixing the CI checks? Also if it's not a lot of trouble, can we try upgrade sqlparse to v0.4.4?

nahuelverdugo:
Sure, I will take a look tomorrow.

nahuelverdugo:
Hi @reata, I have solved the issue with sqlparse and fixed it to `v0.4.4`. There is another issue in one test related to `LATERAL VIEWS` in the latest changes in `sqlfluff`. The error comes from [this](https://github.com/sqlfluff/sqlfluff/pull/4731) PR. 

If you check the PR, instead of a list of column alias ([doc](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-lateral-view.html)), it returns a list of tables. That is causing the error when analyzing this query where `item1` is considered another source table:

```sql
INSERT OVERWRITE TABLE foo
SELECT sc.id, q.item0, q.item1
FROM bar sc
LATERAL VIEW json_tuple(sc.json, 'key1', 'key2') q AS item0, item1
```

We have to open a PR to revert some of those changes.

setup.py
```diff
@@ -65,9 +65,9 @@ def run(self) -> None:
     ],
     python_requires=">=3.8",
     install_requires=[
-        "sqlparse>=0.3.1",
+        "sqlparse==0.4.4",
         "networkx>=2.4",
-        "sqlfluff>=1.4.5",
+        "sqlfluff==2.0.3",
```
reata, line 70: Let's still use `>=` and not pin them to specific version. These breaking change happens rarely.

sqllineage/core/parser/sqlparse/__init__.py
reata, line 1: Since we're doing monkey patching, how about we write a for loop and wrap each patch function in try/except? I have to admit this kind of breaking change from sqlparse can happen later and we don't want to do an release every time for these kind of fix.

reata (changes requested):
Please addressing the comment inline. And I'll do a v1.4.2 release after merging this PR.

reata:
> Hi @reata, I have solved the issue with sqlparse and fixed it to `v0.4.4`. There is another issue in one test related to `LATERAL VIEWS` in the latest changes in `sqlfluff`. The error comes from [this](https://github.com/sqlfluff/sqlfluff/pull/4731) PR.
> 
> If you check the PR, instead of a list of column alias ([doc](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-lateral-view.html)), it returns a list of tables. That is causing the error when analyzing this query where `item1` is considered another source table:
> 
> ```sql
> INSERT OVERWRITE TABLE foo
> SELECT sc.id, q.item0, q.item1
> FROM bar sc
> LATERAL VIEW json_tuple(sc.json, 'key1', 'key2') q AS item0, item1
> ```
> 
> We have to open a PR to revert some of those changes.

So this is why you pin sqlfluff to 2.0.3, correct? If that's the reason we can't use higher sqlfluff version, let's pin it for now then, You can ignore one of my comments inline.

For sqlparse, I'd still prefer use >= 0.4.4 and wrap our monkey patching in try/except.

setup.py
reata: If that `LATERAL VIEW` thing is breaking with sqlfluff>2.0.3, I'm OK pin it to 2.0.3 for now.

nahuelverdugo: Exactly, that is an issue with `sqlfluff>2.0.3`, so we can leave as it is for now.

nahuelverdugo:
> So this is why you pin sqlfluff to 2.0.3, correct? If that's the reason we can't use higher sqlfluff version, let's pin it for now then, You can ignore one of my comments inline.
> 
> For sqlparse, I'd still prefer use >= 0.4.4 and wrap our monkey patching in try/except.

Ok, I will wrap it in a try/except. 

For the `sqlfluff` issue, I will try to open a PR for fixing it this week.

sqllineage/core/parser/sqlparse/__init__.py
nahuelverdugo: @reata, I was thinking about this. I have doubts if adding a try/except when doing the monkey patching will prevent any change that is done in `sqlparse` does not implies an error in `sqllineage`. For example, the `is_keyword` function was moved this time, so the dependencies issues would remain even if we change how we do the monkey patching. 

For me, depending on a specific version of `sqlparse` and `sqlfluff` helps to ensure the results of `sqllineage` will be the same despite the changes done in those libraries. In addition, they can be updated more safely when a new version of `sqllineage` is released. 

What do you think?

reata: I think there're many different cases here:

1) For monkey patching, it is us who's not honoring public interface. When upstream changes and breaks our code, we deserve it. To avoid this, I want to surround this with try/except. 

2) For upstream changing public interface, by intention (with a new major version released) or not (with minor version that 's not supposed to change public API), we react when necessary. This happens rarely and we need to act quick to pin it to last workable version in short term. Do a release so user not impacted. And in the long term switch to >= last workable version.

3) For upstream introducing a bug, it's pretty much same as case 2, only difference it's we rely on upstream team to fix the bug before we switch to >= new bug-free version.

sqlparse regex change is case 1, sqlparse is_keyword call change is case 2, and sqlfluff lateral view issue is case 3.

Generally speaking, we don't want to pin dependency to a specific version. With our case, once we solve the sqlparse ==v0.4.4 issue, we should allow people to use future v0.4.5 once that's released without doing a release ourselves. I can't think of a good example with sqlparse for the moment. Let me give a example with sqlfluff. Say we pin it to 2.0.3, and later sqlfluff releases v2.1.0, introducing a new dialect, say impala, we should allow user to benefit from that without us doing a new release. Hope this clarifies.

reata (changes requested):

reata:
Sorry, I was too busy this week to give feedback. Let me edit the code and do a release ASAP.

sqllineage/core/parser/sqlparse/__init__.py
reata: Sorry for misleading you for the try/except part. Previously I thought the exception has happening during we do patching, turns out it is not thrown until sqlparse parsing the SQL. Then try/except around patching function doesn't make any sense.

reata:
Closes #361

Issue #244, opened 2022-04-06
No Column Lineage Parsed for DML with SELECT query in parenthesis


reata:
SQL:
```sql
INSERT OVERWRITE TABLE tab1
(SELECT col1 FROM tab2)
```

Column lineage is not correctly parsed while table lineage is fine.
```bash
$ sqllineage -f foo.sql -l table 
Statements(#): 1
Source Tables:
    <default>.tab2
Target Tables:
    <default>.tab1

$ sqllineage -f foo.sql -l column
```

The expected column lineage would be
```
<default>.tab1.col1 <- <default>.tab2.col1
```

reata:
Closed by #371

PR #371, opened 2023-05-19
Improve column level lineage for insert & create queries (sqlfluff)

### LINKED ISSUE #212: Specify Column Names in Insert Statement
Hi ~  
"insert into aaa(de00001) select name from a;"      for this sql string expect to be <default>.aaa.de00001<- <default>.a.name
but it turns out to be   <default>.aaa.name <- <default>.a.name, is this something wrong ?

ulixius9:
### Problem statement

In case or paring the queries like 
`create view new_view (c1,c2) as (select col1,col2 from table)`  the table lineage is generated as follows 
```
source tables: [Table: <default>.table]
target tables: [Table: <default>.new_view]
column lineage: []
```
the lineage is accurate but the column level lineage is not generated.

and I just remove the brackets from the sub query `create view new_view (c1,c2) as select col1,col2 from table`, following is the generated result
```
source tables: [Table: <default>.table]
target tables: [Table: <default>.new_view]
column lineage: [(Column: <default>.table.col1, Column: <default>.new_view.col1), (Column: <default>.table.col2, Column: <default>.new_view.col2)]
``` 

now the column level lineage is generated but is incorrect as the column `col1` should have be linked to `c1` of target table and column `col2` should have be linked to `c2` of target table 

the same logic applies to insert into, create table and similar statements.

### Solution

With this pr I'm trying to solve the same problem, when we process the target I save the context of columns involved of target and consider it while generating the column level lineage.

part of fix for https://github.com/open-metadata/OpenMetadata/issues/7427

---
Closes #212 , #244


tests/test_others.py
```diff
@@ -51,7 +51,7 @@ def test_create_as_with_parenthesis_around_table_name():
 
 def test_create_as_with_parenthesis_around_both():
     sql = "CREATE TABLE tab1 AS (SELECT * FROM (tab2))"
-    assert_table_lineage_equal(sql, {"tab2"}, {"tab1"})
+    assert_table_lineage_equal(sql, {"tab2"}, {"tab1"}, skip_graph_validation=True)
```
ulixius9, line 54: since now we are passing the write context of target the subquery itself will not be added into graph as a write node hence the graph of sqlfluff and sqlparse won't match anymore for such cases

tests/test_insert.py
```diff
@@ -23,3 +24,40 @@ def test_insert_into_with_columns_and_select_union():
...
+            (
+                ColumnQualifierTuple("col1", "src_tbl"),
+                ColumnQualifierTuple("random1", "trg_tbl"),
+            ),
+            (
+                ColumnQualifierTuple("col2", "src_tbl"),
+                ColumnQualifierTuple("random2", "trg_tbl"),
+            ),
+        ],
+        test_sqlparse=False,
```
ulixius9, line 44: skipping the sqlparse for all the tests as sqlparse either does not generate or generates incorrect column level lineage as described in pr description

reata (changes requested):
I don't like this idea of adding yet another attribute to SubQueryLineageHolder. Attribute extra_subqueries is added for a quick release of v1.4.x but this doesn't mean this is the correct way. We should at least try put everything in the graph.

We have a dedicated HAS_COLUMN edge type. I was thinking we can add this edge when handling bracketed column_reference, also put index/order into the edge attribute. And later in SubQueryLineageHolder.add_column_lineage, do graph search regarding target table and column, if search hit, we add another handling logic, otherwise fallback to default.

Let me know if you think this is workable.

ulixius9:
@reata adding the query to the graph makes sense to me, but I'm not sure if it should be put as a write node, to me it makes sense to put the query as a read node.

also regarding the `HAS_COLUMN` edge type would that logic also work when the column names are not same?

for example `insert into target_table(col1, col2) as (select c1, c2 from src_table)` this is a valid query and as I've added this in tests as well that `c1` will form lineage with `col1` and `c2` with `col2`.

reata:
> @reata adding the query to the graph makes sense to me, but I'm not sure if it should be put as a write node, to me it makes sense to put the query as a read node in this particular scenario.
> 
> also regarding the `HAS_COLUMN` edge type would that logic also work when the column names are not same?
> 
> for example `insert into target_table(col1, col2) as (select c1, c2 from src_table)` this is a valid query and as I've added this in tests as well that `c1` will form lineage with `col1` and `c2` with `col2`.

Made some edit to my previous comment, SubQueryLineageHolder.add_column_lineage might not be the right place for this.

Here's what my instinct tells me now. Again, this time it could still be wrong. Just for your reference:

1. while handling (col1, col2), we should be able to get target_table from holder.write. We can build the HAS_COLUMN relation of target_table -> col1 and  target_table -> col2. Put the index information as edge attributes so we know col1 is the first column and col2 is the second.
2. wrap the corresponding graph query into a method in class SubQueryLineageHolder, you can even call it target_columns and put a @property decorator, so it still looks like an attribute.
3. everything else is basically just what you do right now.

ulixius9:
@reata make sense let me try it out and get back to you on that!

tests/test_insert.py
```diff
@@ -7,13 +8,19 @@ def test_insert_into():
 
 def test_insert_into_with_columns():
     assert_table_lineage_equal(
-        "INSERT INTO tab1 (col1, col2) SELECT * FROM tab2;", {"tab2"}, {"tab1"}
+        "INSERT INTO tab1 (col1, col2) SELECT * FROM tab2;",
+        {"tab2"},
+        {"tab1"},
+        skip_graph_validation=True,
```
ulixius9, line 14: Had to skip validation for graph as now we are adding the target columns into graph as edge which is not happing for sqlparse yet

sqllineage/core/parser/sqlfluff/extractors/dml_insert_extractor.py
```diff
@@ -97,9 +99,15 @@ def extract(
                     conditional_handler.handle(segment, holder)
 
         # By recursively extracting each subquery of the parent and merge, we're doing Depth-first search
+
+        # we are passing the write context as the subquery for type SUPPORTED_STMT_TYPES (defined above)
+        # would always be a source(read) query and it is important to pass
+        # the write context to generate the correct lineage
         for sq in subqueries:
+            holder.add_read(sq)
```
ulixius9, line 107: added the subquery as read in holder

reata: For top-level subquery in DML like: 1) create table foo as (subquery); 2) insert into foo (subquery), I think the best approach is just to not treat them as subquery by stripping the bracket directly. The code change is pushed.

tests/test_insert.py
reata: Since we're not adding sqlparse support for this feature, I'll just mark test_sqlparse=False and add some comment so we know table-level lineage actually works for this case.

PR #372, opened 2023-05-22
support create table..clone and alter table...swap

### LINKED ISSUE #373: support create table..clone and alter table...swap
CREATE TABLE CLONE:
```sql
create table tab2 CLONE tab1
```
Ref:

- https://cloud.google.com/bigquery/docs/table-clones-create
- https://docs.snowflake.com/en/sql-reference/sql/create-clone

----------

ALTER TABLE SWAP WITH:
```sql
alter table tab1 swap with tab2
```
Ref:

- https://docs.snowflake.com/en/sql-reference/sql/alter-table

ulixius9:
`create table tab2 CLONE tab1` and `alter table tab1 swap with tab2` types of queries are not supported adding the support for the same with this pr

part of fix for https://github.com/open-metadata/OpenMetadata/issues/9389

---
Closes #373

### DIFF
--- a/sqllineage/core/parser/sqlfluff/extractors/ddl_alter_extractor.py
+++ b/sqllineage/core/parser/sqlfluff/extractors/ddl_alter_extractor.py
@@ -13,6 +13,8 @@ class DdlAlterExtractor(LineageHolderExtractor):
     DDL Alter queries lineage extractor
     """
 
+    SOURCE_KEYWORDS = {"EXCHANGE", "SWAP"}
+
     SUPPORTED_STMT_TYPES = [
         "alter_table_statement",
         "rename_statement",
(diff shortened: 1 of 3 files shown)


reata (changes requested):
did some research, turns out BigQuery also supports CREATE TABLE CLONE syntax. Please also add bigquery in the test. Otherwise this PR is ready to be merged.

reata (changes requested):
Besides the inline comment, one thing I forgot to mention yesterday, are you interested in delivering a sqlparse implementation? In v1.4.x, we'll try our best to support both cases. I can do the job if you'd rather spend your time on other issues. Just let me know.

reata:
> @reata I tried with the sqlparse as well but the sqlparse in case of `create table tab2 CLONE tab1` query considers the `clone` as the alias of the table and in case of `alter table tab1 swap with tab2` it considers the `with` keyword as start of an CTE so I'm not sure how to proceed as I don't have much experience with sqlparse, it would be great if you can give me a hand here.

Sure, thanks for the info. Let me try if I can find a way to deal with this.

Issue #377, opened 2023-05-28
Parenthesis confuse parsing of insert/create view


jgeralnik:
Hi!

sqllineage fails to parse the following query:

```
insert into tab3 (with tab1 as (SELECT * FROM tab2) SELECT * FROM tab1);
```

The problem seems to be that `DmlInsertExtractor` does not handle bracketed segments.

I would think that the fix would be something like adding

```
            if segment.type == "bracketed":
                holder |=self.extract(
                    segment,
                    AnalyzerContext(prev_cte=holder.cte, prev_write=holder.write),
                )
```

to `DmlInsertExtractor.extract` but that doesn't work because the segments returned on the next iteration don't contain "with_compound_statement" but instead the individual parts `[<KeywordSegment: ([L:  1, P: 19]) 'with'>, <CTEDefinitionSegment: ([L:  1, P: 24])>,<SelectStatementSegment: ([L:  1, P: 53])>]`.

I would love to hear any other suggestions and would be happy to open a PR myself with some tests if we can find a solution

reata:
This is definitely a bug and I think you're very close to it. 

Here's what my instinct tells me, not guarantee it will work. But you can try after going into "bracketed", based on it starts with CTE or not, you can call DmlCteExtractor or DmlSelectExtractor. Look forward to your PR.

PR #392, opened 2023-06-15
fix missing parse cte in insert

duongphannamhung:
fix issue #377


reata (changes requested):
Please try reuse DmlCteExtractor instead of modifying/appending parsed segments, which makes code logic hard to read for future reference

reata:
Please resolve the conflict.

PR #395, opened 2023-06-19
fix: cte inside bracketed of insert

duongphannamhung:
re-fix issue: #377


reata:
I have removed the `is_cte` function as it doesn't seem that we can reuse it. Also the test case is moved to `test_sqlfluff.py`, which contains all the new features we don't/won't support in sqlparse. Other than that, this PR is perfect fine.

Issue #398, opened 2023-06-22
Incorrect Source Tables recognition when using SqlFluffLineageAnalyzer


Aka-shi:
Sample query
```
sql = """
Insert into target_table

WITH pl AS (
  Select id, school
From table1)

,fb AS (
Select id, school
FROM table2)

SELECT
table3.id, table3.name, table1.school
FROM table3
LEFT JOIN table1 ON table3.id = table1.id

UNION

SELECT
table4.id, table4.name, table2.school
FROM table4
LEFT JOIN table2 ON table4.id = table2.id
"""
```

When I run `result = LineageRunner(sql)` I get the following result which is the correct one. 
```
Result is  Statements(#): 1
Source Tables:
    <default>.table1
    <default>.table2
    <default>.table3
    <default>.table4
Target Tables:
    <default>.target_table
```

But when I provide a dialect `result = LineageRunner(sql, 'ansi')`, I get the following result
```
Result is  Statements(#): 1
Source Tables:
    <default>.table1
    <default>.table2
Target Tables:
    <default>.target_table
```

It doesn't seem to parse the query after CTEs. This looks like a bug. If anyone can guide me on how I can fix this or take this up it would be really helpful since sqlparse is being deprecated in future versions.

reata:
Yes, it's a bug. It's because we didn't handle UNION in [DmlCteExtractor](https://github.com/reata/sqllineage/blob/master/sqllineage/core/parser/sqlfluff/extractors/cte_extractor.py)

UNION is of type `set_expression` in sqlfluff AST, and we should hanlde `set_expression` the same way as `select_statement`. This should be a simple fix, looking forward to your PR.

PR #399, opened 2023-06-25
Fix issue #398: add parse union at the end of cte

duongphannamhung:
Fix issue #398
add test with specific dialect (ansi), work well with Bigquery if we use "UNION ALL" rather than "UNION".

### DIFF
--- a/tests/test_others_dialect_specific.py
+++ b/tests/test_others_dialect_specific.py
@@ -202,3 +202,30 @@ UPDATE SET t.col = s.col"""
         {"tgt"},
         dialect=dialect,
     )
+
+
+@pytest.mark.parametrize("dialect", ["ansi"])
+def test_union_in_insert_with_cte(dialect: str):
+    # issue #398
+    sql = """INSERT INTO target_table
+WITH pl AS (
+  SELECT id, school
+FROM table1)
+, fb AS (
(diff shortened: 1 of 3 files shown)


tests/test_others_dialect_specific.py
reata, line 1: If it's ANSI compatible, then don't put it into dialect specific test suits. `test_cte.py` is the better place for this test case.

tests/test_others_dialect_specific.py
```diff
@@ -202,3 +202,30 @@ def test_merge_insert_row(dialect: str):
...
+FROM table2)
+SELECT
+table3.id, table3.name, table1.school
+FROM table3
+LEFT JOIN table1 ON table3.id = table1.id
+UNION
+SELECT
+table4.id, table4.name, table2.school
+FROM table4
+LEFT JOIN table2 ON table4.id = table2.id"""
```
reata, line 225: If you can simplify the test case, that would be even better. The corner case is with CTE + UNION. We just need one CTE instead of two. Even `INSERT INTO` can be removed.

reata (changes requested):
Please make the change based on the comment inline. Ideally we want each test case to be as atomic as possible.

PR #444, opened 2023-09-04
feat: store original source expression for columns

### LINKED ISSUE #443: Store original expression of every column
This will enhance lineage with how actually columns are calculated. 
Example: 

```sql
SELECT CONCAT(column1, column2) as columnNew FROM table1
```

As of now, for `columnNew`, the lineage produces items for `column1`, `column2` and `*`, but none of them store the original expression `CONCAT(column1, column2) as columnNew`.

vladkol:
Closes #443.

Adds `source_expression` attribute to `Column` which is populated from `BaseSegment.raw` in `SqlFluffColumn.of`.
`Column.to_source_columns` makes sure the attribute is preserved.

### DIFF
--- a/sqllineage/core/holders.py
+++ b/sqllineage/core/holders.py
@@ -304,6 +304,7 @@ class SQLLineageHolder(ColumnLineageMixin):
             for parent in unresolved_col.parent_candidates:
                 src_col = Column(unresolved_col.raw_name)
                 src_col.parent = parent
+                src_col.source_expression = unresolved_col.source_expression
                 if g.has_edge(parent, src_col):
                     src_cols.append(src_col)
             if len(src_cols) == 1:
--- a/sqllineage/core/holders.py
+++ b/sqllineage/core/holders.py
@@ -1,4 +1,3 @@
-import copy
 import itertools
 from typing import Dict, List, Set, Tuple, Union
 
@@ -269,7 +268,7 @@ class SQLLineageHolder(ColumnLineageMixin):
         :class:`sqllineage.holders.SQLLineageHolder`
         """
         g = DiGraph()
-        column_expressions = {}
+        column_expressions: Dict[Tuple[Column, ...], List[str]] = {}
         for holder in args:
             # Storing column expressions
             column_lineage = holder.get_column_lineage(False)
--- a/sqllineage/core/holders.py
+++ b/sqllineage/core/holders.py
@@ -273,8 +273,10 @@ class SQLLineageHolder(ColumnLineageMixin):
             # Storing column expressions
             column_lineage = holder.get_column_lineage(False)
             for item in column_lineage:
-                item_expressions = column_expressions.setdefault(item, [])
-                item_expressions.append(item[0]._source_expression)
+                for i in range(len(item) - 1):
+                    pair = (item[i], item[i + 1])
+                    pair_expressions = column_expressions.setdefault(pair, [])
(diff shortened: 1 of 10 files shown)


reata:
We have similar effort previously and later decide to get it removed. Let me explains the rationale with an example, for a SQL file like
```
INSERT INTO foo
SELECT val + 1 AS val FROM bar;

INSERT INTO bar
SELECT max(val) AS val FROM foo;
```

The column lineage is `bar.val <- foo.val <- bar.val`, Column foo.val is target column in statement 1 while source column in statement 2. Their raw expression is different. However, when merged into graph, foo.val is just one node, since raw_expression is stored as an attribute of Column with same hash id, we lose half the information at graph level. 

With that in mind, I would prefer storing this information in graph edge. So bar.val <- foo.val keeps `val + 1 AS val` and foo.val <- bar.val keeps `max(val) AS val`.

But then we need a friendly API to retrieve it back from graph. That's tricky part to design and I haven't got a good idea yet.

vladkol:
@reata My guess is that you mean the graph it's building for lineage drawing.
My goal is to process lineage programmatically, and I don't seem to lose expression information in that case.
Making it work in drawing would certainly be great, but it requires support in `sqllineagejs` like you mentioned.

vladkol:
Thank you, @reata. I get the problem now. 
I modified my PR by adding column lineage edges support. Please let me know if this is helpful.

reata:
@vladkol thanks for the quick turnaround. Would you mind fix the format? The overall design looks good to me, It'd be nice to add a few unit tests.

At the same time, this is a non-trivial feature, I need to take a deep dive, hopefully on this weekend. (just too much work on weekdays, sorry

reata:
@vladkol Sorry for keep you waiting. I had some really tough time recently, and might still need one or two weeks before getting back to this project.

reata (changes requested):
Sorry for coming back later. Here's a test case that breaks the current implementation:

```python
def test_select_nested_subquery():
    sql = """WITH temp_cte AS (SELECT col0 as col1 FROM tab2)
INSERT INTO tab1
SELECT col1 AS col2
FROM temp_cte"""
    assert_column_lineage_equal(
        sql,
        [(ColumnQualifierTuple("col0", "tab2"), ColumnQualifierTuple("col2", "tab1"))],
    )
    assert_column_lineage_edges_equal(
        sql,
        [
            (
                ColumnQualifierTuple("col0", "tab2"),
                ColumnQualifierTuple("col1", "temp_cte"),
                "col0 as col1",
            ),
            (
                ColumnQualifierTuple("col1", "temp_cte"),
                ColumnQualifierTuple("col2", "tab1"),
                "col1 as col2",
            )
        ],
    )
```

```
Expected Lineage: {(Column: <default>.tab2.col0, Column: temp_cte.col1, 'col0 as col1'), (Column: temp_cte.col1, Column: <default>.tab1.col2, 'col1 as col2')}
Actual Lineage: {(Column: <default>.tab2.col0, Column: temp_cte.col1, 'col0 as col1'), (Column: temp_cte.col1, Column: <default>.tab1.col2, 'col0 as col1')}

```

I think we need to put the raw information into SubQueryLineageHolder instead of Column. Let me know if you're still interested in working on this. I shall have more time available in coming weeks.

reata:
Close as no update. Please create a new PR in case you'd like to continue the work.

PR #456, opened 2023-09-25
[ADD] Support of `DROP VIEW` statements

ovcharenko:
Fixing the exception like

```
sqllineage.exceptions.UnsupportedStatementException: SQLLineage doesn't support analyzing statement type [drop_view_statement] for SQL:DROP VIEW IF EXISTS view1
```


reata (approved):
LGTM

Issue #451, opened 2023-09-12
The problem with the column lineage for repeated aliases


kkozhakin:
Hi! 

There was a problem with building a column lineage due to repeated aliases, in which it is impossible to uniquely correlate paths.

For example, for the query:
```sql
CREATE TEMPORARY TABLE "phone_pd_cur" ON COMMIT DROP AS

SELECT "a1"."phone_pd_id" as "phone_pd_id"
  , "h"."id"
  , "h"."utc_valid_from_dttm"
FROM (
  SELECT t.id
    , t.utc_valid_from_dttm

  FROM (SELECT * FROM "schema1"."h__phone_pd") as t) as h

  LEFT JOIN (SELECT * FROM "schema1"."a__phone_pd__phone_pd_id__key") as "a1"
            ON "a1".id = h.id

DISTRIBUTED RANDOMLY;
ANALYZE "phone_pd_cur";
CREATE TEMPORARY TABLE "email_pd_cur" ON COMMIT DROP AS

SELECT "a1"."email_pd_id" as "email_pd_id"
  , "h"."id"
  , "h"."utc_valid_from_dttm"
FROM (
  SELECT t.id
    , t.utc_valid_from_dttm

  FROM (SELECT * FROM "schema1"."h__email_pd") as t) as h

  LEFT JOIN (SELECT * FROM "schema1"."a__email_pd__email_pd_id__key") as "a1"
            ON "a1".id = h.id

DISTRIBUTED RANDOMLY;
ANALYZE "email_pd_cur";
```
column lineage is:
```
<default>.email_pd_cur.email_pd_id <- a1.email_pd_id
<default>.email_pd_cur.id <- h.id <- t.id
<default>.email_pd_cur.utc_valid_from_dttm <- h.utc_valid_from_dttm <- t.utc_valid_from_dttm
<default>.phone_pd_cur.id <- h.id <- t.id
<default>.phone_pd_cur.phone_pd_id <- a1.phone_pd_id
<default>.phone_pd_cur.utc_valid_from_dttm <- h.utc_valid_from_dttm <- t.utc_valid_from_dttm
a1.* <- schema1.a__email_pd__email_pd_id__key.*
a1.* <- schema1.a__phone_pd__phone_pd_id__key.*
t.* <- schema1.h__email_pd.*
t.* <- schema1.h__phone_pd.*
```

reata:
Can you share the sqllineage version, dialect and expected result?

kkozhakin:
[...]

PR #478, opened 2023-11-07
feat models: change comparing columns

kkozhakin:
feat models: change comparing columns (add parent in compare);
del `.lower()` from __str__ (table names into quotes is case-sensitive)
fixes: #451
test: local

### DIFF
--- a/sqllineage/core/models.py
+++ b/sqllineage/core/models.py
@@ -19,7 +19,7 @@ class Schema:
         self.raw_name = escape_identifier_name(name)
 
     def __str__(self):
-        return self.raw_name.lower()
+        return self.raw_name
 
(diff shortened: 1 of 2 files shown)


reata:
Sorry for the delayed response.

I like the idea of reserving case for table/column name when it's quoted. Always convert to lower case and assume case-insensitive is simply not true for every database.

But on the other hand, different databases handles case differently and not all of them follow ANSI standard. We might need to add that as a global config in the future. As of now, I made some update to `escape_identifier_name` function to make it ANSI compatible. Let me know if that's not working for you.

Thanks for your contribution!

reata (approved):
LGTM

Issue #481, opened 2023-11-24
inconsistent column lineage


maoxingda:
**The results of column lineage using non-validating and Redshift dialects are inconsistent.**

**SQL**

```sql
insert into
    public.t3
(
    c1
)
with
    cte1 as (
        select t1.c1 from public.t1 t1
    ),
    cte2 as (
        select t2.c1 from public.t2 t2
    )
select
    sq1.c1
from
    (
        select cte1.c1 from cte1 union all select cte2.c1 from cte2
    ) sq1
;
```

**To Reproduce**

*Note here we refer to SQL provided in prior step as stored in a file named `test.sql`*

```python
from sqllineage.runner import LineageRunner

with open('test.sql', 'r') as f:
    sql = f.read()

lr1 = LineageRunner(sql)
lr2 = LineageRunner(sql, dialect='redshift')

pprint(lr1.get_column_lineage())
pprint(lr2.get_column_lineage())
```
```
[(Column: public.t1.c1, Column: cte1.c1, Column: sq1.c1, Column: public.t3.c1),
 (Column: public.t2.c1, Column: cte2.c1, Column: sq1.c1, Column: public.t3.c1)]

[(Column: <default>.cte1.c1, Column: sq1.c1, Column: public.t3.c1),
 (Column: <default>.cte2.c1, Column: sq1.c1, Column: public.t3.c1)]
```

**Expected behavior**

```
[(Column: public.t1.c1, Column: cte1.c1, Column: sq1.c1, Column: public.t3.c1),
 (Column: public.t2.c1, Column: cte2.c1, Column: sq1.c1, Column: public.t3.c1)]
```

**Python version (available via `python --version`)**
 - 3.8.18

**SQLLineage version (available via `sqllineage --version`):**
 - 1.4.8

**Additional context**

**lr1._sql_holder.column_lineage_graph visualization**
<img width="1461" alt="non-validating" src="https://github.com/reata/sqllineage/assets/39357378/65a033e6-aca5-4e0a-b863-d0b7f0f3672e">

**lr2._sql_holder.column_lineage_graph visualization**
<img width="1461" alt="redshift" src="https://github.com/reata/sqllineage/assets/39357378/601afc1c-12bf-4053-9641-a81e3d76a78f">

maoxingda:
**Giving aliases t3 and t4 to cte1 and cte2 respectively does not yield the correct results.**

```sql
insert into
    public.t3
(
    c1
)
with
    cte1 as (
[...]

PR #488, opened 2023-12-05
fix CTE and UNION case in #481

maoxingda:
making progress in #481


sqllineage/core/parser/sqlfluff/extractors/select.py
```diff
@@ -150,10 +160,12 @@ def _handle_set(self, segment: BaseSegment) -> None:
                     subquery, alias = sq
                     table_identifier = find_table_identifier(subquery)
                     if table_identifier:
-                        read_sq = SqlFluffTable.of(table_identifier, alias)
                         for seg in list_child_segments(subquery):
+                            for set_subquery in self.list_subquery(seg):
+                                self.subqueries_within_set.append(set_subquery)
+
+                            self._handle_table(seg, holder)
                             self._handle_column(seg)
```
reata, line 168: Let's remove the `list_subqueries` call. We can get children segment of type `select_statement` and `bracketed` from set_expression. And then list_child_segments from `select_statement` and `bracketed` and feed it to `_handle_table` and `_handle_column`.  We don't need subqueries_within_set that way.

tests/sql/column/test_column_cross_stmts_with_similar_subquery_alias.py
reata, line 1: This is a separate issue. Create another PR for this.

reata (changes requested):
Thanks for the PR. Here's the review summary:

1) We don't really need to modify holder implementation. Just try re-use `_handle_table` and `_handle_column` call in `_handle_set` method.
2) Focus on the test case with CTE + UNION on this PR. Only Keep the test case tests/sql/column/test_column_subquery_within_set.py.

sqllineage/core/parser/sqlfluff/extractors/select.py
maoxingda: I don't quite understand. Can you help me by directly modifying my code?

tests/sql/column/test_column_cross_stmts_with_similar_subquery_alias.py
maoxingda: Okay, I will create a separate pull request.

reata:
Code modified and merged. This PR is narrowed to handle CTE+UNION case only. Will handle other cases in original issue in separate PR.

PR #489, opened 2023-12-06
fix: similar alias across statements

maoxingda:


maoxingda:
Hello, greetings! May I inquire if there are plans to merge this pull request?

reata:
For SQL
```sql
insert into
    public.tgt_tbl1
(
    id
)
select
    sq.id
from
    (
        select
            id
        from
            public.src_tbl1
    ) sq

union all

select
    sq.id
from
    (
        select
            id
        from
            public.src_tbl2
    ) sq
```

If the goal is just to make column lineage correct as
```
public.tgt_tbl1.id <- sq.id <- public.src_tbl1.id
public.tgt_tbl1.id <- sq.id <- public.src_tbl2.id
```
We should still go to `SelectExtractor` to update the set expreesion handling logic.


To make it better, if we also want the visualization correct, where we'd to show the first sq is different from the second sq, that sq.id in first line and sq.id in second line are two different column node, then modifying the graph in holder is not the preferred approach.

We should look at `Column` and `SubQuery` data model, leveraging `__hash__` and/or `__str__` method to make sure the node is created correctly when building the graph.

reata:
#490 fixes the `SelectExtractor`, now the column lineage output is correct. Let me know if you're still interested in working on fix the visualization.

reata:
Close as no update. Please create a new PR in case you'd like to continue the work.

PR #509, opened 2023-12-20
feat core/parser: add silent mode to skip not supported statements

### LINKED ISSUE #513: Silent Mode Option to Suppress UnsupportedStatementException
Currently when using any dialect other than non-validating, namely sqlfluff parse implementation, SQLLineage needs to know the statement type before analyzing it. 

For example, SelectExtractor supports statement type of `["select_statement", "set_expression", "bracketed"]`. And we have a special-purpose NoopExtractor to whitelist a bunch of statement types where no lineage can be extracted.

This whitelist approach solves a problem from non-validating dialect, where we don't know what statement we support. And when user reports a exception, it's hard to tell whether it's a bug or feature.

However, this comes at cost that for statement type we haven't included in the whitelist an UnsupportedStatementException will be thrown. Catch the exception is one way. 

In this proposal we'd also like to introduce a new silent_mode option to throw warnings instead of exception.

kkozhakin:
Add silent mode into analizer because it has to be deleted with regular expressions

---
Closes #513

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,9 +103,6 @@ venv.bak/
 # mypy
 .mypy_cache/
 
-# pycharm
-.idea/
-
 # dependencies
 node_modules
 .pnp
@@ -124,3 +121,9 @@ yarn-error.log*
 
 # Optional eslint cache
 .eslintcache
+
+# pycharm
+.idea/
+
+# local testing sql files
+test*.sql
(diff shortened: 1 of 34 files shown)


maoxingda:
Is the silent_mode option also required for the command line?

kkozhakin:
> Is the silent_mode option also required for the command line?

Optional, default is False

reata (changes requested):
I like this silent_mode idea very much. Can you also change the signature of abstract class LineageAnalyzer.analyze to be consistent? Thanks

reata (approved):
LGTM

PR #545, opened 2024-01-10
Set param from config

delphisharp:
Dear ALL：

目前模型的默认模式名是通过环境变量传递的，我认为实现方法不够python。
所以，我实现了另一种解决方法。 
1. 修改 sqllineage/config.py 实现配置项的读取和写入， 读取方法保持和已有方法兼容。
2. 修改 sqllineage/core/models.py 的Schema 类，在每次初始化的时候动态读取配置
3. 修改 sqllineage/runner.py的init方法，传入变量，并在 SQLLineageConfig 赋值
请大家提出意见，谢谢

Currently the default schema name of the model is passed through environment variables, and I think the implementation method is not python enough.
So, I implemented another workaround.
1. Modify sqllineage/config.py to implement reading and writing of configuration items. The reading method remains compatible with the existing method.
2. Modify the Schema class of sqllineage/core/models.py and dynamically read the configuration every time it is initialized.
3. Modify the init method of sqllineage/runner.py, pass in variables, and assign values in SQLLineageConfig
Please give your opinions, thank you

### DIFF
--- a/sqllineage/config.py
+++ b/sqllineage/config.py
@@ -1,28 +1,88 @@
+import dataclasses
 import os
+from dataclasses import dataclass
+from typing import Union, Any
+
+
+@dataclass
+class SQLLineageConfigValue:
+    class_type: Any
+    value: Union[str, bool, None]
+
+
+@dataclass
+class SQLLineageConfigDef:
+    DIRECTORY: SQLLineageConfigValue
+    DEFAULT_SCHEMA: SQLLineageConfigValue
+    TSQL_NO_SEMICOLON: SQLLineageConfigValue
 
 
 class _SQLLineageConfigLoader:
     """
     Load all configurable items from environment variable, otherwise fallback to default
     """
+    def __init__(self):
+        self._config = SQLLineageConfigDef(
+            DIRECTORY=SQLLineageConfigValue(
+                str, os.path.join(os.path.dirname(__file__), "data")
+            ),
+            DEFAULT_SCHEMA=SQLLineageConfigValue(str, None),
+            TSQL_NO_SEMICOLON=SQLLineageConfigValue(bool, False),
+        )
 
-    # inspired by https://github.com/joke2k/django-environ
-    config = {
-        # for frontend directory drawer
-        "DIRECTORY": (str, os.path.join(os.path.dirname(__file__), "data")),
-        # set default schema/database
-        "DEFAULT_SCHEMA": (str, ""),
-        # to enable tsql no semicolon splitter mode
-        "TSQL_NO_SEMICOLON": (bool, False),
-    }
+    @property
+    def DEFAULT_SCHEMA(self):
+        return self._config.DEFAULT_SCHEMA.value
 
-    def __getattr__(self, item):
-        if item in self.config:
-            type_, default = self.config[item]
-            # require SQLLINEAGE_ prefix from environment variable
-            return type_(os.environ.get("SQLLINEAGE_" + item, default))
+    @DEFAULT_SCHEMA.setter
+    def DEFAULT_SCHEMA(self, value):
+        if isinstance(value, self._config.DEFAULT_SCHEMA.class_type):
+            self._config.DEFAULT_SCHEMA.value = value
         else:
-            return super().__getattribute__(item)
+            raise
+
+    @property
+    def DIRECTORY(self):
+        return self._config.DIRECTORY.value
+
+    @DIRECTORY.setter
+    def DIRECTORY(self, value):
+        if isinstance(value, self._config.DIRECTORY.class_type):
+            self._config.DIRECTORY.value = value
+        else:
+            raise
+
+    @property
+    def TSQL_NO_SEMICOLON(self):
+        return self._config.TSQL_NO_SEMICOLON.value
+
+    @TSQL_NO_SEMICOLON.setter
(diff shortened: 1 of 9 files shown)


sqllineage/config.py
```diff
@@ -1,28 +1,81 @@
...
-    }
-
-    def __getattr__(self, item):
-        if item in self.config:
-            type_, default = self.config[item]
-            # require SQLLINEAGE_ prefix from environment variable
-            return type_(os.environ.get("SQLLINEAGE_" + item, default))
-        else:
-            return super().__getattribute__(item)
+    def __init__(self):
```
maoxingda, line 23: mypy不允许方法定义不加 **annotation**

delphisharp: yes . you are  right
see https://mypy.readthedocs.io/en/stable/common_issues.html

change to 
```python
def __init__(self) -> None
``` 
is OK

delphisharp:
已重新提交代码。 本地测试通过。 
Code has been resubmitted. The local pytest passed.

delphisharp:
Due to code format reasons, resubmit.

sqllineage/config.py
```diff
@@ -1,28 +1,77 @@
...
+class SQLLineageConfigValue:
+    class_type: Any
+    value: Union[str, bool, None]
+
+
+@dataclass
+class SQLLineageConfigDef:
+    DIRECTORY: SQLLineageConfigValue
+    DEFAULT_SCHEMA: SQLLineageConfigValue
+    TSQL_NO_SEMICOLON: SQLLineageConfigValue
```
reata, line 16: Make these two dataclass private as well. Only expose instance SQLLineageConfig from this module.

sqllineage/config.py
```diff
@@ -1,28 +1,77 @@
...
-            # require SQLLINEAGE_ prefix from environment variable
-            return type_(os.environ.get("SQLLINEAGE_" + item, default))
+    def __init__(self) -> None:
+        self._config = SQLLineageConfigDef(
+            DIRECTORY=SQLLineageConfigValue(
+                str, os.path.join(os.path.dirname(__file__), "data")
+            ),
+            DEFAULT_SCHEMA=SQLLineageConfigValue(str, None),
+            TSQL_NO_SEMICOLON=SQLLineageConfigValue(bool, False),
+        )
```
reata, line 31: DO NOT make breaking change. Still try read from environment variable first, if not found then fallback to default value.

maoxingda:
https://github.com/maoxingda/sqllineage/blob/0823bf2e32f19d0f64a2c7acb951213db8696b93/sqllineage/config.py#L19

我感觉你这个改动有点问题呀，你现在的写法不支持环境变量的配置了。

I feel there is something wrong with your change. Your current writing method does not support the configuration of environment variables.

https://github.com/reata/sqllineage/pull/540

我正在通过环境变量的方式配置是否启用 **列别名引用功能**

I am configuring whether to enable the **lateral column alias reference function** through environment variables.

sqllineage/core/models.py
```diff
@@ -5,19 +5,20 @@
 from sqllineage.exceptions import SQLLineageException
 from sqllineage.utils.helpers import escape_identifier_name
 
+schema_unknown = "<default>"
```
reata, line 8: Any particular reason that you'd like to move this variable out of Schema?

sqllineage/runner.py
```diff
@@ -42,6 +42,8 @@ def __init__(
         verbose: bool = False,
         silent_mode: bool = False,
         draw_options: Optional[Dict[str, str]] = None,
+        default_schema: Optional[str] = None,
+        tsql_no_semicolon: Optional[bool] = False,
```
reata, line 46: I'd prefer explicitly ask user to set value on SQLLineageConfig, and not through LineageRunner interface.

reata (changes requested):
Please address the comment inline. Also check the uncovered lines and add unit test for it.

reata:
> https://github.com/maoxingda/sqllineage/blob/0823bf2e32f19d0f64a2c7acb951213db8696b93/sqllineage/config.py#L19
> 
> 我感觉你这个改动有点问题呀，你现在的写法不支持环境变量的配置了。
> 
> I feel there is something wrong with your change. Your current writing method does not support the configuration of environment variables.
> 
> #540
> 
> 我正在通过环境变量的方式配置是否启用 **列别名引用功能**
> 
> I am configuring whether to enable the **lateral column alias reference function** through environment variables.

Agreed, we need to maintain support config through environment variables. This approach is read only.

The new ability we're adding through this PR is to set config at run time via Python code.

delphisharp:
刚才在添加测试用例，各位意见现在看到了。我总结下：
1.  优先走环境变量
2. 单独从配置类中读取
3. 参数不通过Runner传递，通过设定 config类实现


I was adding test cases just now, and your comments are now visible. Let me summarize:
1. Prioritize environment variables
2. Read from the configuration from config Class 
3.  he parameters are not passed through the Runner，  set config class

delphisharp:
use black,  
```python
item[self._len_prefix : len(self._PREFIX)] 
```
must be have whitespace.

but use flask8,
./sqllineage/config.py:38:47: E203 whitespace before ':'
            self._config[item[self._len_prefix : len(self._PREFIX)]][


how to do?

PR #552, opened 2024-01-14
fix: SQLLineageConfig boolean value returns True for all non-empty strings (#551)

### LINKED ISSUE #551: SQLLineageConfig boolean value returns True for all non-empty strings
**SQLLineageConfig** returns True for all non-empty strings

**Expected**
```
'true', 'on', 'ok', 'y', 'yes', '1'
```

For the strings in the above list, return True, otherwise return False

maoxingda:
fix: SQLLineageConfig returns True for all non-empty strings (#551)

Closes #551

### DIFF
--- a/sqllineage/config.py
+++ b/sqllineage/config.py
@@ -15,14 +15,39 @@ class _SQLLineageConfigLoader:
         # to enable tsql no semicolon splitter mode
         "TSQL_NO_SEMICOLON": (bool, False),
     }
+    BOOLEAN_TRUE_STRINGS = ("true", "on", "ok", "y", "yes", "1")
 
     def __getattr__(self, item):
         if item in self.config:
             type_, default = self.config[item]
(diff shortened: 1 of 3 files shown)


sqllineage/config.py
```diff
@@ -15,14 +15,39 @@ class _SQLLineageConfigLoader:
...
+    def parse_value(cls, value, cast):
+        """Parse and cast provided value
+
+        :param value: Stringed value.
+        :param cast: Type to cast return value as.
+
+        :returns: Casted value
+        """
+        if cast is None:
+            return value
```
reata, line 40: I don't think this logic is needed for existing configs, any particular reason we want to add this?

maoxingda:
Agree, I copied it from another project, can you help me remove it directly?

reata:
> Agree, I copied it from another project, can you help me remove it directly?

sure

PR #553, opened 2024-01-15
fix: Set param from config

delphisharp:
Dear ALL：

目前模型的默认模式名是通过环境变量传递的，我认为实现方法不够python。
所以，我实现了另一种解决方法。

修改 sqllineage/config.py 实现配置项的读取和写入， 读取方法保持和已有方法兼容。
修改 sqllineage/core/models.py 的Schema 类，在每次初始化的时候动态读取配置
修改 sqllineage/runner.py的init方法，传入变量，并在 SQLLineageConfig 赋值
请大家提出意见，谢谢
Currently the default schema name of the model is passed through environment variables, and I think the implementation method is not python enough.
So, I implemented another workaround.

Modify sqllineage/config.py to implement reading and writing of configuration items. The reading method remains compatible with the existing method.
Modify the Schema class of sqllineage/core/models.py and dynamically read the configuration every time it is initialized.
Modify the init method of sqllineage/runner.py, pass in variables, and assign values in SQLLineageConfig
Please give your opinions, thank you

重新建立分支并提交
Re-branch and submit

first step to solve #536

### DIFF
--- a/sqllineage/config.py
+++ b/sqllineage/config.py
@@ -30,7 +30,9 @@ class _SQLLineageConfigLoader:
 
     def __getattr__(self, item: str):
         if item in self.config.keys():
-            if value := self._thread_config.get(self.get_ident(), {}).get(item, None):
+            if (
+                value := self._thread_config.get(self.get_ident(), {}).get(item)
+            ) is not None:
                 return value
 
             type_, default = self.config[item]
@@ -81,14 +83,13 @@ class _SQLLineageConfigLoader:
                     value, self.config[key][0]
                 )
             else:
-                super().__setattr__(key, value)
+                raise ConfigException(f"Invalid config key: {key}")
         return self
 
     def __enter__(self):
         if self._in_context_manager:
             raise ConfigException("SQLLineageConfig context manager is not reentrant")
         self._in_context_manager = True
-        return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         if self.get_ident() in self._thread_config:
--- a/sqllineage/config.py
+++ b/sqllineage/config.py
@@ -2,6 +2,8 @@ import os
 import threading
 from typing import Any, Dict
 
+from sqllineage.exceptions import ConfigException
+
 
 class _SQLLineageConfigLoader:
     """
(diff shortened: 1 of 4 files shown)


reata:
Do we have all the codes in this new branch? I remember we need to change the Schema model in #545

delphisharp:
> Do we have all the codes in this new branch? I remember we need to change the Schema model in #545

Sorry, there were some problems when my local branch was synchronized from upstream changes of the master. 
So a new branch was reopened.

reata:
What's the difference between this and #555 ? Do they jointly solve #536 and is there any dependency, like I need to merge this PR first and then #555?

sqllineage/config.py
```diff
@@ -3,32 +3,36 @@
...
+                "class_type": str,
+                "additional": None,
+                "default": None,
+            },
+            "TSQL_NO_SEMICOLON": {
+                "class_type": bool,
+                "additional": None,
+                "default": False,
+            },
+        }
```
reata, line 28: Let's not change the config data structure. Just create a separate variable to store the value set at runtime.

reata (changes requested):
Change config at runtime is a dangerous thing as we discussed previously under multi-thread context. Although we normally encourage user to use multi-process for bulk process, but thread-safety is still something we need to guarantee given SQLLineageConfig is a global variable. 

With that in mind, I'd like to propose this API leveraging Python context manager:

```python
with SQLLineageConfig.set(DEFAULT_SCHEMA="ods", TSQL_NO_SEMICOLON=True):
    runner = LineageRunner(sql)
    print(runner.source_tables)
```
Once entering the with block, those config are temporarily set to another value for current thread and are only readable to current thread. And the temporary value is cleaned up after exit with block.

At the heart of the implementation, we should have a data structure like:
```
{
    "thread_1": {"DEFAULT_SCHEMA": "ods", "TSQL_NO_SEMICOLON": True},
    "thread_2": {"DEFAULT_SCHEMA": "default"}
}
```
which takes highest priority > environment variable > default value

reata (changes requested):
Please stick to our design that user need to explicitly call set method instead of \_\_setattr\_\_:

```python
with SQLLineageConfig.set(DEFAULT_SCHEMA="ods", TSQL_NO_SEMICOLON=True):
    runner = LineageRunner(sql)
    print(runner.source_tables)
```

reata (approved):
LGTM

PR #601, opened 2024-04-24
Update sqlalchemy.py

zhengtingxue:
add support get field list from hive metastore mysql database


zhengtingxue:
add support get field list from hive metastore mysql database

delphisharp:
Hello.
I would like to know what is the purpose of this pull?

Based on typical scenarios, I have a few questions:
insert into target_table partition(month_id='202301',day_id='22') select *** from source_tab 

1. Should the collected field list be stored in field order?
    I understand that field order is necessary for subsequent use of the metadata. When column information of target_table is required, the field order determines the mapping relationship of source_table columns.

2. Is it necessary to eliminate or distinguish whether the field is a partition field?
    The partition field should not be used as a column in the target_table or at least needs to be distinguished

3. It is recommended to use f-string writing method for string instead of splicing

zhengtingxue:
Thank you for your reply. I realize that I have not considered enough, and my thoughts were too simplistic.

PR #633, opened 2024-06-23
delete some intermediate tables

gyafdxis-k:
Whether we try to delete some intermediate tables. Deletion using regular matching, such as tables starting with ^tmp can be deleted in the blood relationship. The upstream and downstream of tables starting with ^tmp can be directly connected according to the Cartesian product.

### DIFF
--- a/sqllineage/core/holders.py
+++ b/sqllineage/core/holders.py
@@ -379,6 +379,7 @@ class SQLLineageHolder(ColumnLineageMixin):
         metadata_provider: MetaDataProvider, *args: StatementLineageHolder
     ) -> DiGraph:
         g = DiGraph()
+        g1 = DiGraph()
         for holder in args:
             g = nx.compose(g, holder.graph)
             if holder.drop:
--- a/sqllineage/core/metadata_provider.py
+++ b/sqllineage/core/metadata_provider.py
@@ -60,6 +60,17 @@ class MetaDataProvider:
         bool value tells whether this provider is ready to provide metadata
         """
         return True
+        
+    @classmethod
+    def str2enum(cls, name):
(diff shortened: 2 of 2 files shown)


gyafdxis-k:
del some intermediate tables function but I think this feature can be config

delphisharp:
tmp or temp is a business definition.
 I don't think it's a good idea to introduce business definitions. 
Business processing should be done at the business level.
What if my temporary table definition is based on the schema name? Or other logic?

delphisharp:
Is there any problem in processing business logic on the parsing results?
I mean, a tool should have its own boundaries, so I don't think this business processing should be done at the tool level.

If you want to do it, you should use an interface or other more compatible methods.
Scenarios I encountered
1. Create Session Table in teradata
2. Schema is Tmp or ods_temp, is temp table
In this case, I process the parsing results at the business level.

For the current metadata_provider, I also think that providing a callback function is enough, and how to implement it is up to the developer to handle it.

gyafdxis-k:
> mp or temp is a business definition.
> I don't think it's a good idea to introduce business definitions.
> Business processing should be done at the business level.
> What if my temporary table definition is based on the schema name? Or other logic?

It is only a example, I think the callback function is a good idea which is friendly to the user
thanks for your replying, I think the sqlineage is very good opensource project

reata:
Agreed with @delphisharp . It's better for user to post-process the lineage results generated by sqllineage. 

^tmp would be a very simple scenario for temporary table naming convention. I know cases like some company organizes data teams in a decentralized way, and each data team has its own database for temporary table (and production table). It would be almost impossible if we're targeting solving this temporary table problem.

Issue #644, opened 2024-08-13
Upgrade `sqlfluff` dependency to 3.1.0+


KulykDmytro:
**Describe the solution you'd like**
Upgrade `sqlfluff` dependency to 3.1.0+
There are many needed fixes happened

shoffmeister:
FWIW, I just built (and only built!) `sqllineage` from source, with the `sqlfluff` dependency set to 3.1.1 - all tests are passing, without any further change in sqllineage.

This would suggest that locally you could simply run `sqllineage` in a virtual environment of its own, and then simply upgrade the `sqlfluff` dependency using `pip`?

(Granted a new release of `sqllineage` with latest `sqlfluff` would be more convenient, but then `sqlfluff` seems to move at an astonishing pace, which I can imagine makes it challenging to track that)

shoffmeister:
FWIW, I verified that running
```
pip install --upgrade https://github.com/sqlfluff/sqlfluff/tarball/master
```
in "my" sqllineage Python virtual environment did complete to success (I needed sqlfluff master)

I also got the graph and the output that I wanted out of that hot-patched installation of sqllineage.

Your mileage may vary.

PR #660, opened 2024-11-15
feat: upgrade sqlfluff to 3.2.x

rubytobi:
Closes https://github.com/reata/sqllineage/issues/644

Since [SQLFluff 3.2.0](https://github.com/sqlfluff/sqlfluff/releases/tag/3.2.0), "the bracketed arguments are now wrapped in a `function_contents` object". I adjusted the code base so that it works again.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,7 +67,7 @@ setup(
     install_requires=[
         "sqlparse==0.5.0",
         "networkx>=2.4",
-        "sqlfluff~=3.2.5",
+        "sqlfluff==3.3.0",
         "sqlalchemy>=2.0.0",
     ],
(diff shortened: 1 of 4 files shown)


reata:
Sorry for the delayed response, been struggling in my work for past months. 

This PR makes all the sense to me. I made a few edit though:

1. upgrade sqlfluff to 3.3.0, as this is the newest release and I see no new breaking changes.
2. pin the sqlfluff version to 3.3.0, not using tilde equal because we rely on a lot of implementation details in sqlfluff. And based on our past experience, those can change within minor versions. To mitigate, I'll try release sqllineage on a monthly basis this year (if my workload allows for it) so that we can stay up to date with sqlfluff.
3. split a dedicated logic for function type handling to make it a bit more readable.

Thanks for your contribution!

PR #685, opened 2025-02-08
feat: improve performance on large SQL/graphs

rubytobi:
As discussed in https://github.com/reata/sqllineage/discussions/679, I've done a first draft of minimal changes to speed up the column level lineage identification.

I think we should keep `networkx` - it's more powerful and flexible - and just switch to `rustworkx` where needed for performance reason.

Keen to hear your thoughts!

### DIFF
--- a/docs/gear_up/configuration.rst
+++ b/docs/gear_up/configuration.rst
@@ -90,6 +90,24 @@ Default: ``False``
 
 Since: 1.4.8
 
+GRAPH_OPERATOR_CLASS
+====================
+A rustworkx graph operator implementation is available since v1.5.7. Set this config to
+``sqllineage.core.graph.rustworkx.RustworkXGraphOperator`` to enable it for better performance, in particular
+when handling large lineage graphs.
+
+SQLLineage uses networkx as default graph operator implementation. Any import errors when using other graph operator
+will fallback to networkx implementation.
+
+.. note::
+     This is an experimental feature. Based on our test suite, rustworkx graph operator shows significant performance
+     improvement over networkx implementation and passes all tests. However, there might be edge cases not covered
+     by our tests. Please report any issues you encounter when using rustworkx graph operator.
+
+Default: ``sqllineage.core.graph.networkx.NetworkXGraphOperator``
+
+Since: 1.5.7
+
 
 .. _Amazon Redshift announces support for lateral column alias reference: https://aws.amazon.com/about-aws/whats-new/2018/08/amazon-redshift-announces-support-for-lateral-column-alias-reference/
 .. _Support "lateral column alias references" to allow column aliases to be used within SELECT clauses: https://issues.apache.org/jira/browse/SPARK-27561
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,9 +6,10 @@ build-backend = "hatchling.build"
 name = "sqllineage"
 dependencies = [
     "sqlparse>=0.5.4",
-    "networkx>=2.4",
     "sqlfluff>=3.2.0",
     "sqlalchemy>=2.0.0",
+    "networkx>=2.4",
+    "rustworkx>=0.16.0",
 ]
 requires-python = ">=3.10"
 authors = [
--- a/sqllineage/config.py
+++ b/sqllineage/config.py
@@ -20,6 +20,11 @@ class _SQLLineageConfigLoader:
         "TSQL_NO_SEMICOLON": (bool, False),
         # lateral column alias reference supported by some dialect (redshift, spark 3.4+, etc)
         "LATERAL_COLUMN_ALIAS_REFERENCE": (bool, False),
+        # graph operator implementation class
+        "GRAPH_OPERATOR_CLASS": (
+            str,
+            "sqllineage.core.graph.networkx.NetworkXGraphOperator",
+        ),
     }
 
     def __init__(self) -> None:
--- a/sqllineage/core/graph/__init__.py
+++ b/sqllineage/core/graph/__init__.py
@@ -1,0 +1,26 @@
+import importlib
+import logging
+import typing
+
+from sqllineage.config import SQLLineageConfig
+from sqllineage.core.graph.networkx import NetworkXGraphOperator
+
+logger = logging.getLogger(__name__)
+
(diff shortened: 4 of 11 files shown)


reata (commented):
Can you write a performance test to show that rustworkx can handle large graph within certain time compared to networkx?

Also if we're converting networkx graph to rustworkx, then I don't think adding a graph property is needed. You can directly make a copy from self._sql_holder.graph.

rubytobi:
I can't share the graph, i.e. nodes and edges itself due to confidentiality. However, I've ran a test locally using `pytest-benchmark` with attached results.

image

Do you have some complex queries available for using as a benchmark?

reata:
I didn't mean for sharing your sql query. Any test case added should **NOT** be related to real business. 

Instead, we should be able to build a hypothetical SQL file, abstracting away the business, while still be able to reproduce the complexity of the graph, that justifies the introduction of rustworkx.

rubytobi:
image

I've added a simple test case The benefits grow the larger the graph is.

First row represents the commit in this state: https://github.com/reata/sqllineage/blob/9a2f33c330080a8d0a67f13d95a790bf9c6e34c6/sqllineage/core/holders.py

reata (changes requested):
I was thinking of something more simpler. Fake a large SQL ourselves, and then run using timeit. Supposing the original networkx implementation takes 100s, and rustworkx takes 10s, and we can safely assert the running time less than 50s. A naive example:

```python
def test_performance():
    from timeit import timeit

    def _test_performance():
        sql = "\n".join([f"insert into tab{i + 1} select col from tab{i};" for i in range(10)])
        runner = LineageRunner(sql=sql, dialect="ansi")
        return runner.get_column_lineage()

    assert timeit(_test_performance, number=10) < 3
```
Note this fake SQL doesn't show advantages of rustworkx over networkx though. Both takes ~3.5s in my machine. Maybe we need a more clever approach to build this sql to reflect the complexity of the graph. But this is the rough idea.

We don't need to introduce pytest-benchmark for this purpose. (pytest-benchmark only show benchmark numbers, not guarding any test case running duration within a time threshold, correct?)

rubytobi:
I looked into `pytest-benchmark` because of it's calibration and robustness:

> `pytest-benchmark` will run your function multiple times between measurements. A _round_ is that set of runs done between measurements. This is quite similar to the builtin `timeit` module but it’s more robust. 
[pytest-benchmark/calibration](https://pytest-benchmark.readthedocs.io/en/latest/calibration.html)

However, I think your approach works just as well. I've updated the code accordingly.

I noticed tho, your SQL and mine are quite different and thus perform differently as well. You are benchmarking parsing the SQL and building the internal graph while the changes purely affect establishing the lineage while traversing the graph.

I found a good stackoverflow comment going into the arguments for [timeit](https://stackoverflow.com/questions/48258008/n-and-r-arguments-to-ipythons-timeit-magic/59543135#59543135).

```sql
insert into tab1 select col from tab0;
insert into tab2 select col from tab1;
insert into tab3 select col from tab2;
insert into tab4 select col from tab3;
insert into tab5 select col from tab4;
insert into tab6 select col from tab5;
insert into tab7 select col from tab6;
insert into tab8 select col from tab7;
insert into tab9 select col from tab8;
insert into tab10 select col from tab9;
```
* on networkx: `<0.6`
* on rustworkx: `<0.5`


```sql
CREATE TABLE database_b.table_b
  SELECT col_a,
         col_b,
         col_c,
         col_d,
         col_e,
         col_f,
         col_g,
         col_i,
         col_j,
         col_k
  FROM   database_a.table_a
```
* on networkx: `<0.5`
* on rustworkx: `<0.2`

reata:
Sorry for the inconvenience as I'm trying to modernizing this project with pyprojec.toml. As of now, please use ruff as linter only and we will still stick with black as formatter. You can refer to the following section for CI command to be executed:
```
[tool.tox.env_run_base]
deps = [".[ci]"]
commands = [
    ["black", ".", "--check", "--diff"],
    ["ruff", "check", "."],
    ["bandit", "-r", "sqllineage"],
    ["mypy"],
    ["pytest", "--cov"],
]
```

As for pytest-benchmark or timeit, I'm both OK. I think the key and most difficult part is to have a SQL that shows the complexity so that performance gap between networkx and rustworkx is so big that we can use a unit test for guard it. On the other hand, if it's not big enough, then when changing to a different machine, the test might not stand still.

rubytobi:
@reata, I've reworked the proposed changes. 

Rather than switching completely to `networkx`, I've implemented `rustworkx` as opt-in solution, i.e. you must call `runner.get_column_lineage` with `use_rustworkx=True`. This doesn't have to be the final solution, but it's the easiest to demonstrate and test the performance gains.

Additionally, I've made the test case more complex, i.e. the `rustworkx` based solution is forced to be at least twice as fast as `networkx` across all Python and OS versions used on GitHub Actions.

I switched to [pytest-speed](https://github.com/samuelcolvin/pytest-speed) which allows to interact with the benchmark results. `pytest-benchmark` relies on persisting benchmark results as json files as reference over time.

```
Statements(#): 8
Source Tables:
    database_a.table_a
    database_a.table_b
Target Tables:
    database_b.table_d
    database_c.table_a
    database_d.table_b
Intermediate Tables:
    database_b.table_a
    database_b.table_b
    database_b.table_c
    database_c.table_b
    database_d.table_a
```

rubytobi:
Hi @reata, I'd still be keen to get this merged :)

reata:
@rubytobi Sorry, been stuck in work for weeks.

The current implementation is much better, and I like the idea of opt-in. But I still can't help thinking that this is a partial solution, as we only let people opt-in when calling get_column_lineage. Allow me a few more weeks to think on a more satisfactory way to introduce rustworkx. (the value is undeniable, thanks for pointing that out)

rubytobi:
No worries, @reata 

Given the performance improvement, I would suggest removing the networkx part and only keep the rustworkx implementation for this section.

networkx still has a place in constructing the graph. It's simple and straight forward to use in this case. Just when it comes to analysing the established graph we should switch to rustworkx.

reata (approved):

PR #710, opened 2025-08-03
chore: upgrade sqlfluff to 3.4.2

morningman:
3.4.2 support new SQL dialect: Apache Doris and Flink SQL


reata:
fixed a breaking change introduced by sqlfluff 3.4.2 on tsql assignment operator

Issue #643, opened 2024-08-08
Parse stored procedures using sqllineage


godhao123:
Really love sqllineage! I have a question: I want to parse SQL statements from stored procedures. Is there a good method to extract valid SQL statements from stored procedures?

kendn1993:
me too

maoxingda:
Preprocess the contents of the stored procedure before using SQLLineage to analyze the lineage.

godhao123:
> Preprocess the contents of the stored procedure before using SQLLineage to analyze the lineage.

Could you please advise on how to effectively preprocess stored procedures, such as removing irrelevant syntax keywords and extracting complete, valid SQL?

maoxingda:
- My own approach is as follows:

  - Standardize the writing of stored procedures (limit certain writing methods, such as supporting three levels of nested judgment statements, etc.)

  - Use regular expressions to extract grammatically correct SQL, and then use sqllineage to parse the lineage

godhao123:
> * My own approach is as follows:
>   
>   * Standardize the writing of stored procedures (limit certain writing methods, such as supporting three levels of nested judgment statements, etc.)
>   * Use regular expressions to extract grammatically correct SQL, and then use sqllineage to parse the lineage

Thank you very much for your response. I will strive to do better.

kendn1993:
I use sqlparse to parse procedures.

reata:
Close via #767 with preliminary support for bigquery, oracle, and tsql

thanks @AlexJBSilva for the contribution.

PR #761, opened 2026-03-23
feat: support parsing stored procedures for bigquery dialect

AlexJBSilva:
Address (in part, and specifically for bigquery dialect) what was asked in Discussion #556 and Issue #643.

The task is accomplished by adding the case for `multi_statement_segment` in
method `_list_specific_statement_segment` from class `SqlFluffLineageAnalyzer`
combined with the method `recursive_crawl("statement")` to extract all statements
from multi statement segments and a filter for processing only the statement types
supported by sqllineage (no warnings or exceptions as procedures can contain
many statements meaningless for lineage construction).

To cope with many statements inside a multi statement segment, the parse and
cache strategy from `analyzer.split_tsql()` method was leveraged (probably this
method should be renamed as it seems this approach can be useful for parsing
procedures from other dialects).

This pull request is in draft mode as I just made parsing stored procedures for bigquery dialect to work.
Any thoughts are welcome for getting this PR ready for merge.

### DIFF
--- a/sqllineage/core/parser/sqlfluff/analyzer.py
+++ b/sqllineage/core/parser/sqlfluff/analyzer.py
@@ -91,9 +91,18 @@ class SqlFluffLineageAnalyzer(LineageAnalyzer):
                 f"{sql}\n"
                 f"{violation_msg}"
             )
+        supported_types = []
+        for extractor_cls in BaseExtractor.__subclasses__():
+            supported_types.extend(extractor_cls.SUPPORTED_STMT_TYPES)
         segments = []
         for top_segment in getattr(parsed.tree, "segments", []):
             match top_segment.type:
+                case "multi_statement_segment":
+                    for statement in top_segment.segments[0].recursive_crawl(
+                        "statement"
+                    ):
+                        if statement.segments[0].type in supported_types:
+                            segments.append(statement.segments[0])
                 case "statement":
                     segments.append(top_segment.segments[0])
                 case "batch":
--- a/sqllineage/runner.py
+++ b/sqllineage/runner.py
@@ -190,7 +190,11 @@ Target Tables:
                 self._file_path, self._dialect, self._silent_mode
             )
         )
-        if SQLLineageConfig.TSQL_NO_SEMICOLON and self._dialect == "tsql":
+        if (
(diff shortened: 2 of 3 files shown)


AlexJBSilva:
This is still a Work in Progress as the test is passing with sqlfluff 4.0.4 but failing with sqlfluff 4.1.0.
My implementation was relying on `CREATE PROCEDURE` statement being wrapped by sqlfluff bigquery dialect exclusive `multi_statement_segment` type. But the wrapping changed to the common `statement` type in https://github.com/sqlfluff/sqlfluff/pull/7534.

reata:
Hi @AlexJBSilva , thanks for contributing. Sorry it took me a while.

Here's some early feedback:

1. For sqlfluff behavior change, this is not new. We're relying the internal details of sqlfluff, so it's a tax we pay. To get this moving, we can upgrade sqlfluff to >=4.1.0, and then implement this feature using the new parsed result. Actually I think `statement` makes sense than `multi_statement_segment`.
2. I don't like we share the logic with TSQL_NO_SEMICOLON. As documented in https://sqllineage.readthedocs.io/en/latest/gear_up/configuration.html#tsql-no-semicolon, this config is to be removed in the long term, the `split_tsql` will be gone too. 

Putting implementation effort aside, the way I'm thinking about stored procedure is, should we split one create procedure statement into multiple SQL statement, or we tackle it within a single statement? Personally I'd prefer the latter approach, but let me know your thought.

AlexJBSilva:
Hi @reata.
Sorry it also took me a while to answer you back.

1. I agree that `statement` makes more sense for `CREATE PROCEDURE` than `multi_statement_segment`, especially because this was the behavior for other dialects.
But there's no need to limit sqllineage to `sqlfluff to >=4.1.0`.
It should be enough to inform BigQuery users that if they want the new procedure parsing feature, they need to have `sqlfluff to >=4.1.0`.

2. I reused the split_tsql() because it was the **easy shortcut** to get procedure parsing to work.

Following the architecture of sqllineage, processing a `create procedure statement` should definitely be implemented within a single statement.
This demands a higher effort into learning the inner workings of sqllineage, but the reward is that we can support `create procedure statements` for other dialects (including bigquery).

Considering this broader scope, I am closing this PR in favor of PR #767.

PR #767, opened 2026-05-01
feat: support parsing stored procedures

AlexJBSilva:
Address what was asked in Discussion #556 and Issue #643.

Add ProcedureExtractor to parse CREATE PROCEDURE statements.
Extracts table and column lineage from procedure body by:
- Using recursive_crawl to find nested statements
- Delegating to existing extractors (Select, Insert, etc.)

**Supports BigQuery, Oracle and T-SQL dialects.**

Tests added:
- tests/sql/table/test_procedure_dialect_specific.py
- tests/sql/column/test_column_procedure_dialect_specific.py

---

Currently, SQLFluff version 4.1.0 has some degree of support for parsing stored procedures in the following dialects: bigquery, mariadb, mysql, oracle, postgres, redshift, snowflake and tsql.
Among those dialects, only bigquery, oracle, and tsql parsed trees have enough elements for some lineage extraction.

- Dialects `mariadb` and `mysql` are **not** supported because SQLFluff can't cope with dynamic delimiters (see https://github.com/sqlfluff/sqlfluff/issues/7081);
- Dialects `postgres` and `redshift` are **not** supported because SQLFluff treats the procedure body (internals) as a single string literal (`quoted_literal` type);
- Dialect `snowflake` is **not** supported because SQLFluff treats the procedure body (internals) as a single string literal (`udf_body` type).

### DIFF
--- a/sqllineage/core/parser/sqlfluff/extractors/procedure.py
+++ b/sqllineage/core/parser/sqlfluff/extractors/procedure.py
@@ -1,4 +1,3 @@
-import textwrap
 import warnings
 
 from sqlfluff.core.parser import BaseSegment
@@ -34,14 +33,8 @@ class ProcedureExtractor(BaseExtractor):
             if extractor.can_extract(segment.type):
                 return extractor.extract(segment, context)
         else:
-            if not self._silent_mode:
-                char_limit = 50
-                segment_text = textwrap.shorten(
-                    segment.raw, width=char_limit, placeholder=" ..."
-                )
-                warnings.warn(
-                    "SQLLineage doesn't support analyzing statement type "
-                    f"[{segment.type}] for SQL Segment: '{segment_text}'"
-                )
-                return SubQueryLineageHolder()
+            warnings.warn(
+                "SQLLineage doesn't support analyzing statement type "
+                f"[{segment.type}] for SQL Segment: '{segment.raw}' embedding in procedure, skipping it."
+            )
             return SubQueryLineageHolder()
--- a/sqllineage/core/parser/sqlfluff/extractors/procedure.py
+++ b/sqllineage/core/parser/sqlfluff/extractors/procedure.py
@@ -28,10 +28,11 @@ class ProcedureExtractor(BaseExtractor):
     def _delegate_to_extractor(
         self, segment: BaseSegment, context: AnalyzerContext
     ) -> SubQueryLineageHolder:
-        for extractor_cls in BaseExtractor.__subclasses__():
-            extractor = extractor_cls(self.dialect, self.metadata_provider)
-            if extractor.can_extract(segment.type):
-                return extractor.extract(segment, context)
+        holder = BaseExtractor.try_extract(
+            self.dialect, self.metadata_provider, segment, context
+        )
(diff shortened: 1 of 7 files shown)


sqllineage/core/parser/sqlfluff/extractors/procedure.py
```diff
@@ -0,0 +1,33 @@
...
+            holder |= self._delegate_to_extractor(segment.segments[0], context)
+        return holder
+
+    def _delegate_to_extractor(
+        self, segment: BaseSegment, context: AnalyzerContext
+    ) -> SubQueryLineageHolder:
+        for extractor_cls in BaseExtractor.__subclasses__():
+            extractor = extractor_cls(self.dialect, self.metadata_provider)
+            if extractor.can_extract(segment.type):
+                return extractor.extract(segment, context)
```
reata: SqlFluffLineageAnalyzer has a similar logic and it raise an UnsupportedStatementException (unless in silent mode).

For stored procedure, I think it makes sense to keep running for unsupported statement, but let's create a warnings.warn to make debug easier.

Also I was thinking since both share similar logic, maybe we should move this logic to `BaseExtractor`, make both `SqlFluffLineageAnalyzer` and `ProcedureExtractor` to re-use it. But this is a nice-to-have.

tests/sql/table/test_procedure_dialect_specific.py
```diff
@@ -0,0 +1,69 @@
...
+        sql,
+        {"tab1"},
+        {"tab2"},
+        dialect=dialect,
+        test_sqlparse=False,
+    )
+
+
+@pytest.mark.parametrize("dialect", proc_dialects)
+def test_procedure_column_multiple_statements(dialect: str):
```
reata: maybe this should be called `test_procedure_multiple_statements`?

reata (changes requested):
The core idea is solid.

The parameterized tests that use `{'' if dialect == 'bigquery' else 'AS'}` inline makes the SQL hard to read. It less desirable but I don't know a better way to do it. So I guess it is what it is.

Two minor issues to be addressed and I think we should be good to merge.

sqllineage/core/parser/sqlfluff/extractors/procedure.py
AlexJBSilva: Implemented `warnings.warn` in `_delegate_to_extractor` to make debugging easier.
The applied behavior is to always show a warning when there is no extractor capable of extracting the statement segment (unsupported statement) and no warnings when in silent_mode.

From a user experience perspective, as procedures can contain many statements meaningless for lineage extraction, I think that an alternative approach would be to show warnings only if a `debug flag` is on, and no warnings by default.

Deferred moving the shared logic to `BaseExtractor` for a future iteration.

reata: I changed the code to just fall back to silent mode regardlessly for statements within procedure. No need to pass it down.

Also I created a `try_extract` method in `BaseExtractor` to unify the two.

reata (approved):
LGTM. 

Close #643

PR #778, opened 2026-05-25
perf: reduce hot-path overhead in column lineage analysis

rubytobi:
# perf: reduce hot-path overhead in column lineage analysis

Profiling a 1000+ statement production workload identified four hot spots in the sqllineage analysis phase (~44s of a ~335s total). A fifth change is ergonomic. 

Claude was used during this work. All existing tests pass, new tests for identified behaviour not yet tested.

---

### Batch SQLFluff parsing

Every `Linter(config=...)` construction deepcopies `FluffConfig`. With 1000+ statements, that ran 1000+ times. A new `preparse(statements)` on `SqlFluffLineageAnalyzer` joins all statements and calls `_list_specific_statement_segment` once, caching each `(stmt, segment)` pair in `_segment_cache`. `analyze()` hits the cache; on a segment-count mismatch it falls back to per-statement parsing transparently. `LineageRunner._eval` calls `preparse` before the statement loop (non-TSQL sqlfluff path only). Cache keys are original input strings to ensure hits regardless of trailing-semicolon differences.

Note: The combined SQL string may hit SQLFluff's default `rust_parser_max_iterations`/`max_parse_nodes` limits. Is that something you'd want to make expose via `SQLLineageConfig`?

### Hoist loop-invariant call in `end_of_query_cleanup`

`holder.get_alias_mapping_from_table_group(tbl_grp)` (full edge scan for `HAS_ALIAS`) was called once per column despite `tbl_grp` being loop-invariant. Moved before the outer loop.

### O(1) tag index for `retrieve_vertices_by_props`

This method (called 176k times in profiling) previously scanned every node on every call. Both `NetworkXGraphOperator` and `RustworkXGraphOperator` now maintain a `_tag_index: dict[str, set]`, updated in `add_vertex_if_not_exist`, `update_vertices`, `drop_vertices`, and `merge`. The index stores the canonical vertex (first object added for a logical vertex) to prevent broken lineage chains from equal-but-distinct `SubQuery` objects in CTE lookups.

### Cache `Table.__str__` and `Column.__str__`; optimise `Column.parent`

`Table.__str__` (12.1M calls) is now built once at `__init__`. `Column.__str__` (8.6M calls) uses a lazy `_str_cache` invalidated by the `parent` setter. `Column.parent` replaces `next(iter(self._parent))` with a for-loop early return to avoid iterator allocation.

### Accept `list[str]` as `sql` in `LineageRunner`

Callers holding statements as a list no longer need to join them for `LineageRunner` to split again. Ergonomic only; the `split()` call is negligible relative to sqlfluff parsing.

---

## Benchmark

`benchmarks/bench_performance.py` (included).

```
  Micro: Table.__str__ / Column.__str__ (with parent)         1.7x / 2.9x

  retrieve_vertices_by_props  O(N) scan vs O(1) index
  500 nodes   NetworkX  23x    RustworkX  56x
  2000 nodes  NetworkX  40x    RustworkX  96x

  Batch parse vs sequential
  60 stmts   2.05x      120 stmts  1.96x      480 stmts  1.81x
```

The tag-index speedup scales with graph size (O(1) vs O(N)). RustworkX gains more because its raw node iteration is slower. Batch parse speedup decreases at higher statement counts as SQLFluff's internal cost grows non-linearly relative to the fixed deepcopy overhead.


rubytobi:
@reata, would love to work with you to get these improvements into the code base.

reata:
> @reata, would love to work with you to get these improvements into the code base.

Thanks, amazing work! The core idea looks solid. I'll take some time to review the details and possibly come back with questions. Meanwhile if you can take care of the coverage drop, that would be great.

rubytobi:
@reata, we found a bug in the hoisting of `write_columns`. I've reverted the specific changes and added an extra test to capture the behaviour. If you think that's actually incorrect behaviour, we can get the change back in.

reata:
@rubytobi  Sorry, it's been an eventful month. I'm starting to review piece by piece now. Here's the feedback regarding **Batch SQLFluff parsing**. 

I know it's been a problem as we made the attempt to unify what sqlparse and sqlfluff do on parsing. sqlfluff implementation was awkwardly following the split and the parse each statement approach, which isn't desirable compared to its own batch parsing capability.

However, the original PR's batch-parse optimization introduced a design problem in `runner.py`:

```python
# runner._eval — the runner knows about concrete analyzer types
if isinstance(analyzer, SqlFluffLineageAnalyzer) and not (...):
    analyzer.preparse(self._stmt)
```

This violates the OOP design. `LineageAnalyzer` is an abstract base class defining an `analyze()` contract, but the runner reaches into a concrete subclass to call `preparse()` — a method the base class knows nothing about. It also forces the runner to import `SQLLineageConfig` and handle TSQL-specific branching that should be the analyzer's concern. The runner was doing too many jobs: splitting SQL, deciding whether to batch-parse, and only then delegating to `analyze()`.

The PR also changed the public API **Accept list[str] as sql in LineageRunner** — a change I felt was unnecessary for an internal optimization.

So I appended cfab72a603f3be08def6f754414f93e2f4a9c00f to this PR, to summarize:

**Made the analyzer own its input end-to-end.** The base class now takes `sql` in its constructor and declares an abstract `statements` property. Each analyzer is responsible for its own splitting and pre-processing:

```python
class LineageAnalyzer(ABC):
    def __init__(self, sql: str, **_):
        self._sql = sql

    @property
    @abstractmethod
    def statements(self) -> list[str]: ...

    @abstractmethod
    def analyze(self, sql: str, metadata_provider) -> StatementLineageHolder: ...
```

**SqlFluffLineageAnalyzer** batch-parses eagerly in its constructor via `_split_and_cache()`. The batch parse happens once — one `Linter()` call instead of N — and the resulting segments are cached keyed by statement string. The `statements` property returns the cached list.

**SqlParseLineageAnalyzer** splits on demand in its `statements` property using `sqlparse.parse()`.

**Runner._eval** just asks the analyzer for `statements`, then loops:

```python
self._stmt = analyzer.statements
with session:
    for stmt in self._stmt:
        holder = analyzer.analyze(stmt, session.metadata_provider)
        ...
```

Since I'm changing this part of files, I also:
- Eliminated the `split()` → rejoin → reparse roundtrip in `_split_and_cache`. Sqlfluff handles multi-statement SQL natively — no need to pre-split with sqlparse.
- Merged `split_tsql` into `_split_and_cache`. The "batch" case in `_list_specific_statement_segment` already iterates child statements regardless of `TSQL_NO_SEMICOLON`. The dedicated TSQL code path was redundant.
- Inlined `split()` and `trim_comment()` from `helpers.py` into `SqlParseLineageAnalyzer`. These were never general-purpose utilities — they only served the sqlparse backend. Their TODO comments ("make parser independent") are now moot.
- Reverted the `list[str]` public API change on `LineageRunner.__init__`.

Unlike your original implementation which tried to keep backward compatible with a fallback to sqlparse split when sqlfluff parses different number of statement with sqlparse, these changes introduce potential breaking changes:

1. **For sqlfluff-backed dialects, statement splitting now goes through sqlfluff instead of sqlparse.** The original code on master always used `helpers.split()` (sqlparse-based) for all dialects. In the refactored code, `SqlFluffLineageAnalyzer` uses sqlfluff's own parser to split statements. 

2. **`TSQL_NO_SEMICOLON` is functionally dead.** The config key remains in `config.py` for backward compatibility, but no code path reads it anymore. The sqlfluff analyzer handles multi-statement SQL (with or without semicolons) uniformly through its `_list_specific_statement_segment` method's "batch" case. 

But consider we do want sqlfluff dialect to parse/split the raw sql, and TSQL_NO_SEMICOLON was flagged as deprecated anytime previously, I think it makes sense to make this change.

reata:
**Hoist loop-invariant call in end_of_query_cleanup** is also good for merge.

However, after close review. I'm having doubt regarding **O(1) tag index for `retrieve_vertices_by_props`** and **Cache `Table.__str__` and `Column.__str__`; optimise `Column.parent`**

### The profiling gap

A high call count alone doesn't make a bottleneck. What matters is **self time as a percentage of total runtime**. A method called 12 million times at 10ns each costs 0.12s. A method called 100 times at 100ms each costs 10s. Without sampling data showing the **actual wall-clock time spent inside** these methods, the call counts and micro-benchmark speedups don't tell us whether the optimizations matter.

| Hotspot         | Evidence given                                                               | What's missing                 |
| --------------- | ---------------------------------------------------------------------------- | ------------------------------ |
| Batch parsing   | N `FluffConfig` deepcopies → 1. Reasoning is solid even without timing data. | —                              |
| Tag index       | 176k calls, 23x–96x micro-benchmark speedup                                  | Self time / % of total runtime |
| `__str__` cache | 12.1M + 8.6M calls, 1.7x–2.9x micro-benchmark speedup                        | Self time / % of total runtime |

For batch parsing, the structural argument is strong regardless of the exact number: replacing N `deepcopy(FluffConfig)` calls with one is unambiguously correct. The tag index and `__str__` caching don't have the same structural clarity. 

I'd suggest revert the tag index (hotspot 3) and `__str__` caching (hotspot 4). If profiling data shows these methods have non-trivial self-time, they can be re-submitted as standalone PRs with cleaner implementations. Hotspots 1 and 2 are good to go. Let me know your thought. @rubytobi

rubytobi:
@reata thanks for looking into these! I've reverted as per your suggestion the `__str__` caching and the tag index work, also dropped the benchmark script for a clean repo state. I left the test cases for behavioural nuances identified during the work here to prevent regressions.

I'll look into these two hotspots from our analysis again and bring new PRs if still appropriate.

reata:
thanks, @rubytobi 

Also I found a potential performance downgrade introduced by my commit https://github.com/reata/sqllineage/commit/cfab72a603f3be08def6f754414f93e2f4a9c00f when dealing with `unload_statement`. But looks like now your PR is coming from an organization fork instead of personal fork, the `Maintainers are allowed to edit this pull request` doesn't work anymore. So I created a subsequent PR (#793) to address that. Just FYI.

reata:
BTW, I saw your new PR on **Cache Table.__str__ and Column.__str__**. I still don't get it why `__str__` can become bottleneck but I'll dig deep.

Meanwhile, regarding **O(1) tag index for retrieve_vertices_by_props**, it occurs to me the other day, that what you did is sort of like creating an index for a database/table. Without that, query node by property requires O(N) full graph scan. 

Despite rustworkx/networkx being a graph library without index-like capability, I did find similar mechanism in graph database:
- neo4j: https://neo4j.com/docs/cypher-manual/current/indexes/search-performance-indexes/create-indexes/#create-a-single-property-range-index-for-nodes
- janusgraph: https://docs.janusgraph.org/schema/index-management/index-performance/#property-indexes

So conceptual-wise, it makes perfect sense. We'll just have to make it look like a graph index abstraction. I don't have anything concrete in my mind now, just want to share this thought.

rubytobi:
I moved the fork to an org so I can use CodSpeed's GitHub Action runners for more consistent wall time benchmark measurements than on my laptop, to much noise locally.

PR #788, opened 2026-07-06
feat: add string enum for SQL dialects

bontail:


reata (commented):
Thanks for this PR — clean implementation, tests pass, backward compat is intact, and the compat layer for Python <3.11 is clean.

One architectural concern: the `Dialect` enum hardcodes a snapshot of sqlfluff's dialect list, but sqlfluff is the real source of truth (`SUPPORTED_DIALECTS` is populated at runtime via `dialect_readout()`). Since we depend on `sqlfluff>=3.2.0` (unpinned), a user can upgrade sqlfluff, get new dialects, but then have their type checker complain because the enum hasn't caught up. So the enum introduces a coupling between the sqllineage release cycle and the sqlfluff release cycle that doesn't exist today.

May I understand the rationale of introducing StrEnum? For typo detection, I believe sqlfluff always emits friendly error message. For auto complete, that's a gain but not worth the long term maintenance effort as flagged above. Happy to know if there's anything I'm missing.

PR #792, opened 2026-07-16
perf: cache Table/Column __str__ and __hash__ to avoid repeated recomputation

rubytobi:
## Summary

`Table` and `Column` are hashed and stringified very frequently during lineage resolution, in set membership checks, graph node lookups, and deduplication, yet `__str__` and `__hash__` recomputed their result from scratch on every single call. Profiling large workloads showed this recomputation as a significant, avoidable cost, since the underlying data these methods depend on rarely, if ever, changes after construction.

This PR caches `Table.__str__`/`__hash__` and `Column.__str__`/`__hash__` instead of recomputing them each time. `Table`'s cache is computed once in `__init__`, since `schema` and `raw_name` never change afterward. `Column`'s cache is populated lazily on first access and invalidated whenever `Column.parent` is set, since `parent` is assigned after construction and can change. The result is a broad, low risk speedup with no behavior change.

## Performance

Custom benchmark suite ([bench.py](https://github.com/user-attachments/files/30086218/bench.py)), before versus after:

| Benchmark | Before (s) | After (s) | Delta |
| --- | --- | --- | --- |
| wide | 60.531 | 23.985 | reduced 60.39% |
| many | 7.949 | 4.375 | reduced 44.97% |
| athena | 114.095 | 96.972 | reduced 15.01% |

With `wide` and `many` being pathological tests, and `athena` being an internal production scenario. [CodSpeed](https://app.codspeed.io/tobycraft/sqllineage/runs/compare/6a58bd500547a28bc90f53e7..6a588a2e7355afc16259ca3f?uri=benchmarks%2Fcodspeed%2Ftest_tpcds_print_column_lineage.py%3A%3Atest_tpcds_print_column_lineage&runnerMode=WallTime
)  (`benchmarks/codspeed/test_tpcds_print_column_lineage.py`):

- `test_tpcds_print_column_lineage`: 192.6s to 130.7s, improved 47%
- `test_tpcds_print_column_lineage_rustworkx`: 94.2s to 92.2s, improved 2%

The networkx variant gains far more because `nx.DiGraph` stores its adjacency as a plain Python dict keyed by the `Table`/`Column` objects themselves, so nearly every graph operation calls `__hash__`/`__str__` on them repeatedly. The rustworkx backend only keeps a thin vertex-to-index Python dict; once a vertex has an index, traversal and edge storage happen inside the Rust core on integer indices, so there is far less redundant hashing for this change to remove.

### DIFF
--- a/sqllineage/core/models.py
+++ b/sqllineage/core/models.py
@@ -63,9 +63,11 @@ class Table:
             self.schema = schema
             self.raw_name = name if escaped else escape_identifier_name(name)
         self.alias = escape_identifier_name(kwargs.pop("alias", self.raw_name))
+        self._str_cache = f"{self.schema}.{self.raw_name}"
+        self._hash_cache = hash(self._str_cache)
 
     def __str__(self):
-        return f"{self.schema}.{self.raw_name}"
+        return self._str_cache
 
     def __repr__(self):
         return "Table: " + str(self)
@@ -74,7 +76,7 @@ class Table:
         return isinstance(other, Table) and str(self) == str(other)
 
     def __hash__(self):
-        return hash(str(self))
+        return self._hash_cache
 
     @staticmethod
     def of(table: Any) -> "Table":
@@ -167,11 +169,17 @@ class Column:
         self.from_alias = kwargs.pop("from_alias", False)
 
     def __str__(self):
-        return (
-            f"{self.parent}.{self.raw_name}"
-            if self.parent is not None and not isinstance(self.parent, Path)
-            else f"{self.raw_name}"
-        )
+        try:
+            return self._str_cache
+        except AttributeError:
+            p = self.parent
+            result = (
+                f"{p}.{self.raw_name}"
+                if p is not None and not isinstance(p, Path)
+                else self.raw_name
+            )
+            self._str_cache = result
+            return result
 
     def __repr__(self):
         return "Column: " + str(self)
@@ -184,7 +192,11 @@ class Column:
         )
 
     def __hash__(self):
-        return hash(str(self))
+        try:
+            return self._hash_cache
+        except AttributeError:
+            result = self._hash_cache = hash(str(self))
+            return result
 
     @property
     def parent(self) -> Path | Table | SubQuery | None:
@@ -193,6 +205,8 @@ class Column:
     @parent.setter
     def parent(self, value: Path | Table | SubQuery):
         self._parent.add(value)
+        self.__dict__.pop("_str_cache", None)
+        self.__dict__.pop("_hash_cache", None)
 
     @property
     def parent_candidates(self) -> list[Path | Table | SubQuery]:


reata:
The benchmark result is unquestionable. For my own education purposes, keep my research notes here.

I changed the `bench.py` a bit to run against two graph libs in one shot:

```python
def run(name: str, statements: int, columns: int, graph_lib: str) -> None:
    with SQLLineageConfig(GRAPH_OPERATOR_CLASS=graph_lib):
        sql = generate_sql(statements, columns)
        runner = LineageRunner(sql, dialect="ansi")
        t0 = time.perf_counter()
        _ = runner.source_tables  # forces parse + analysis + merge
        t1 = time.perf_counter()
        runner.get_column_lineage()
        t2 = time.perf_counter()
        print(
            f"{name} - {graph_lib.rsplit('.')[-1]}: analyze={t1 - t0:.3f}s  column_lineage={t2 - t1:.3f}s"
        )
        print(f"  checksum: {lineage_checksum(runner)}")


if __name__ == "__main__":
    for graph_lib in [
        "sqllineage.core.graph.networkx.NetworkXGraphOperator",
        "sqllineage.core.graph.rustworkx.RustworkXGraphOperator",
    ]:
        run("wide (3 statements x 600 columns)", 3, 600, graph_lib)
        run("many (100 statements x 5 columns)", 100, 5, graph_lib)

```

It shows similar-level of performance boost (~63% reduction) for both networkx and rustworkx on my local, 
comparing PR-792 to master branch. 
```
(venv) ➜  sqllineage git:(master) ✗ python bench.py
wide (3 statements x 600 columns) - NetworkXGraphOperator: analyze=1.964s  column_lineage=52.178s
  checksum: 0229b15e29de (cols=1800)
many (100 statements x 5 columns) - NetworkXGraphOperator: analyze=1.127s  column_lineage=4.130s
  checksum: f51053a1684c (cols=500)
wide (3 statements x 600 columns) - RustworkXGraphOperator: analyze=2.037s  column_lineage=5.672s
  checksum: 0229b15e29de (cols=1800)
many (100 statements x 5 columns) - RustworkXGraphOperator: analyze=0.517s  column_lineage=0.448s
  checksum: f51053a1684c (cols=500)
(venv) ➜  sqllineage git:(master) ✗ gh pr checkout 792
Switched to branch 'perf/str-hash-cache'
(venv) ➜  sqllineage git:(perf/str-hash-cache) ✗ python bench.py   
wide (3 statements x 600 columns) - NetworkXGraphOperator: analyze=2.361s  column_lineage=19.012s
  checksum: 0229b15e29de (cols=1800)
many (100 statements x 5 columns) - NetworkXGraphOperator: analyze=2.019s  column_lineage=1.506s
  checksum: f51053a1684c (cols=500)
wide (3 statements x 600 columns) - RustworkXGraphOperator: analyze=2.829s  column_lineage=2.104s
  checksum: 0229b15e29de (cols=1800)
many (100 statements x 5 columns) - RustworkXGraphOperator: analyze=1.840s  column_lineage=0.162s
  checksum: f51053a1684c (cols=500)
```

I also did the same test on tpcds using `bench.py` by only change the sql to be all files in `sqllineage/data/tpcds` 
concatenated. And networkx bootst is ~65% reduction while rustworkx is ~54% reduction.

---

This makes me a bit unsure. If networkx and rustworkx get similar level of performance boost as my local test suggests, we should probably merge this PR as is.

But on the other hand, if your theory was correct, I was thinking maybe we should change `NetworkXGraphOperator` in how we use networkx, possibly by copying `RustworkXGraphOperator` with vertex-to-index dict.
> The networkx variant gains far more because nx.DiGraph stores its adjacency as a plain Python dict keyed by the Table/Column objects themselves, so nearly every graph operation calls __hash__/__str__ on them repeatedly. The rustworkx backend only keeps a thin vertex-to-index Python dict; once a vertex has an index, traversal and edge storage happen inside the Rust core on integer indices, so there is far less redundant hashing for this change to remove.

I would love to know your thought before we proceed as there might be things I'm unaware of on CodSpeed side setup. @rubytobi

rubytobi:
I agree on merging this change here as-is. It's a small change with identical outcomes, just faster.

The CodSpeed benchmark for TPCDS using rustworkx shows only a 2% wall time improvement because it's dominated by the remaining sqlfluff parsing time. sqlfluff are working on a rust parsing backend with some exciting performance gains btw.

I don't feel strongly about copying rustworkx’s vertex-to-index dict into NetworkXGraphOperator. From quick profiling, I'm not sure how much we'd actually gain. However, we'd lose some of the simplicity in the networkx implementation by maintaining a mapping. If someone needs performance, they can go with the rustworkx backend.

----

One of the performance improvement possibilities we see internally is slightly reframing and extending the question sqllineage currently answers; "What is the full lineage of all tables and columns in these SQL statements?" This requires print_column_lineage to find paths  between all sources and targets.

Internally, and maybe other users of sqllineage too, we are far more often interested in a very specific subset of lineage, e.g. "what's the lineage involving this specific node/identifier?"
Currently we have to compute the full lineage and then filter out. Filtering is cheap but the upfront  effort is huge, especially for large and complex SQLs. Being able to specify single/multiple identifiers and effectively reduce the number of nodes to consider would bring a significant speed up for us. But that's clearly beyond the scope of this PR. Happy to do a prototype to demonstrate what I mean and we can see if that's something fitting into sqllineage or for us to keep separate.

reata:
> it's dominated by the remaining sqlfluff parsing time.

That explains. I was only counting column_lineage time and analyze time was kept out. So yes, if cache `__str__` and `__hash__` can benefit both graph backend, we should merge as is. And I agree that it's unnecessary to make further investment on networkx backend performance.

---

As for subset of lineage, I think I get the idea but I don't have a clue on how to achieve that with sqllineage. Looking forward to that POC.

reata (approved):
LGTM

PR #794, opened 2026-07-19
fix: unquote tsql bracket identifier containing a quote char

crow-crypto-crow:
escape_identifier_name checked for quote chars before the tsql square-bracket branch, so a bracketed identifier containing a quote (e.g. [O'Brien]) matched the quote branch and kept its brackets. Check brackets first. Adds a regression test.


reata (approved):
LGTM. Thanks for your contribution!

PR #795, opened 2026-07-19
fix: capture subquery in scalar projection and CASE ELSE branch

crow-crypto-crow:
list_subqueries only handled CASE WHEN/THEN and function-wrapped subqueries, dropping the source table of a scalar sub-select used directly as a projected column and of a subquery in a CASE ELSE branch. Both are now collected, matching the sqlparse parser. Adds two table-level regression tests.


reata (approved):
Very clear intention and clean implementation!

PR #796, opened 2026-07-19
fix: keep column lineage for tsql UPDATE/MERGE when SET RHS is wrapped in expression

crow-crypto-crow:
In tsql the right-hand side of an assignment is wrapped in an `expression` segment, so set_clause exposes only one direct column_reference and the len==2 guard skipped the source->target edge. Unwrap a single-column-reference expression RHS in both UpdateExtractor and MergeExtractor, with regression tests.


reata (approved):
LGTM

PR #797, opened 2026-07-19
fix: use path containment check instead of string prefix in drawing server

crow-crypto-crow:
The POST d/f parameter guard used str.startswith, which accepts a sibling directory that merely shares the root's string prefix. Use Path.resolve().is_relative_to for a real containment check. Adds a regression test.


reata (approved):
LGTM

PR #798, opened 2026-07-19
fix: return all edges from RustworkXGraphOperator.retrieve_edges_by_vertex without label filter

crow-crypto-crow:
The loop reused the method's own `label` filter parameter as a per-edge local, so when called with label=None only edges sharing the first edge's label were returned. This diverged from NetworkXGraphOperator and dropped lineage across RENAME. Use a dedicated local and add a parametrized regression test over both operators.

### DIFF
--- a/sqllineage/core/graph/rustworkx.py
+++ b/sqllineage/core/graph/rustworkx.py
@@ -149,14 +149,17 @@ class RustworkXGraphOperator(GraphOperator):
                 tgt_vertex = (
                     tgt_node.get("vertex") if isinstance(tgt_node, dict) else tgt_node
                 )
-                label = edge_data["label"]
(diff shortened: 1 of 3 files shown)


reata (changes requested):
The fix is straight-forward, while the bug is a bit difficult to understand, especially on its impact to lineage analysis.

Let's construct a SQL test instead of `test_graph_operator.py`. I found the following test can reproduce this bug, we can add it to `tests/sql/table/multiple_statements/test_tmp_table.py`

```python
def test_alter_source_table_name():
    assert_table_lineage_equal(
        "INSERT INTO tab1 SELECT * FROM tab2; ALTER TABLE tab2 RENAME TO tab3;",
        {"tab3"},
        {"tab1"},
    )
```

reata (approved):
Test case updated

PR #799, opened 2026-07-19
docs: fix column lineage examples and verbose output order

crow-crypto-crow:
Correct the MetaData example (foo.col2 comes from baz not bar; quux has only col5/col6; add the missing print_column_lineage call), fix the undefined sq alias in the column example, and swap the verbose table drop/rename lines to match actual output. Examples now reproduce verbatim.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,14 +76,14 @@ Statement #1: insert into db1.table1 select * from db2.table2;
     table read: [Table: db2.table2]
     table write: [Table: db1.table1]
     table cte: []
-    table rename: []
     table drop: []
+    table rename: []
 Statement #2: insert into db3.table3 select * from db1.table1;
     table read: [Table: db1.table1]
     table write: [Table: db3.table3]
     table cte: []
-    table rename: []
     table drop: []
+    table rename: []
 ==========
 Summary:
 Statements(#): 2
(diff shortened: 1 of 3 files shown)


docs/first_steps/advanced_usage.rst
```diff
@@ -145,15 +145,15 @@ table `quux` has columns `col5` and `col6` and `baz` has column `col4`.
 .. code-block:: bash
 
     sqlite3 db.db 'CREATE TABLE IF NOT EXISTS baz (bar_id int, col1 int, col4 int)';
-    sqlite3 db.db 'CREATE TABLE IF NOT EXISTS quux (quux_id int, col5 int, col6 int)';
+    sqlite3 db.db 'CREATE TABLE IF NOT EXISTS quux (col5 int, col6 int)';
```
reata: No need to remove quux_id here. The intention was every table has its own PK column and that PK column doesn't have to be present in the column lineage when not selected.

README.md
```diff
@@ -177,14 +177,14 @@ Suppose all the tables are created in sqlite database with a file called `db.db`
 table `quux` has columns `col5` and `col6` and `baz` has column `col4`. 
 ```shell
 sqlite3 db.db 'CREATE TABLE IF NOT EXISTS baz (bar_id int, col1 int, col4 int)';
-sqlite3 db.db 'CREATE TABLE IF NOT EXISTS quux (quux_id int, col5 int, col6 int)';
+sqlite3 db.db 'CREATE TABLE IF NOT EXISTS quux (col5 int, col6 int)';
```
reata: Same here.

reata (changes requested):
Thanks for fixing the doc. Minor suggestion to be addressed and we're good to go.

reata (approved):
updated, ready to merge

PR #803, opened 2026-08-07
POC: node-scoped column lineage

rubytobi:
Following up the idea from [here](https://github.com/reata/sqllineage/pull/792#issuecomment-5083032813).

## Motivation

`print_column_lineage()` always computes the full column lineage: every source/target pair goes through `all_simple_paths`, which is the combinatorially expensive part of `get_column_lineage`. Often we only care about lineage touching one specific table, column, or identifier.

Filtering the full result afterward saves nothing, since the expensive path enumeration has already run. This POC pushes the filter into the enumeration stage itself, so only the paths that actually pass through the requested node are computed.

## What has been implemented?

Adds an optional `node` parameter to `LineageRunner.get_column_lineage()` and `print_column_lineage()`:

```python
LineageRunner(sql).print_column_lineage(node="tab2.col1")
```

- `node` accepts a `str`, `Column`, or `Table` identifier and must resolve to exactly one column vertex. Multiple matches raise a new `AmbiguousNode` exception (carrying `.matches`); zero matches returns an empty result.
- Internally, `ColumnLineageMixin.get_column_lineage` computes the ancestor and descendant subgraphs of the resolved node via BFS, then runs `all_simple_paths` only within those subgraphs (source to node, node to target), instead of over the full source x target product.
- `node=None` leaves the existing code path untouched.
- Verified by an equivalence test against naive full-lineage filtering, plus a test asserting significantly fewer `list_lineage_paths` calls when a node filter is applied.

This is a library-only POC on `get_column_lineage`/`print_column_lineage`, deliberately kept small to validate the idea. It does not touch:

- Parsing or graph construction (`_eval()`), which still runs in full; the speedup is limited to path enumeration. I don't think we can realistically extend this to `_eval()`.
- Multiple nodes in a single call
- A `direction` option (upstream/downstream only) and a CLI flag.
- and probably more...


sqllineage/core/holders.py
```diff
@@ -35,23 +87,99 @@ def get_column_lineage(
...
+    def _cone_paths(
+        self, seed: Column, source_columns: set[Column], target_columns: set[Column]
+    ) -> list[list[Column]]:
+        """
+        Path enumeration restricted to the subgraph around ``seed``: only
+        ancestors/descendants of ``seed`` are considered, so path enumeration
+        runs on a much smaller subgraph than the full lineage graph.
+        """
+        ancestors = self._lineage_neighbors(seed, EdgeDirection.IN)
+        descendants = self._lineage_neighbors(seed, EdgeDirection.OUT)
```
reata: We have similar ideas in frontend app that when a column node is clicked, all its ancestors and descendants are highlighted (https://github.com/reata/sqllineage/blob/master/sqllineagejs/src/features/editor/DAG.jsx#L85). In cytoscape.js, these two methods are called `predecessors` and `successors`.

I'm pretty sure networkx and rustworkx has similar built-in functions that are more efficient than our `self._lineage_neighbors`. We should probably let `GraphOperator` interface to provide these two methods and only do union then get_sub_graph here. 

The core idea is that this BFS logic belongs to `GraphOperator`, not `LineageHolder`

sqllineage/core/holders.py
```diff
@@ -4,27 +4,79 @@
...
+                and c not in matches
+            ]
+        else:
+            col_name, table_tail = parts[-1], parts[:-1]
+            for c in columns:
+                if c.raw_name != col_name or c.parent is None:
+                    continue
+                parent_parts = str(c.parent).split(".")
+                if parent_parts[-len(table_tail) :] == table_tail:
+                    matches.append(c)
```
reata: Generally I don't like this str match both column and table logic. It's error prone. `Table("tab")` and "tab" actually return different things and it's confusing for user.

If Python API is sufficient, we can drop the else branch. But if CLI is a must, consider rephrase it like a keyword-based graph search API. Don't mix it with node-based API.

sqllineage/core/holders.py
```diff
@@ -35,23 +87,99 @@ def get_column_lineage(
...
+            for source in true_sources - {seed}
+            for path in up_graph.list_lineage_paths(source, seed)
+        ]
+        down_paths = [[seed]] if seed in true_targets else []
+        down_paths += [
+            path
+            for target in true_targets - {seed}
+            for path in down_graph.list_lineage_paths(seed, target)
+        ]
+        return [up + down[1:] for up in up_paths for down in down_paths]
```
reata: In edge case, when a node is both source and target as a self-loop. This produces a 1-element path. We actually met this issue before with rustworkx upgrade, see https://github.com/Qiskit/rustworkx/issues/1617. 

So networks allows for 1-element path, while rustworkx always generates [seed, seed] as path. This is yet another reason we want to hide it under `GraphOperator` interface. 

And please make sure self-loop (both table and column level) is included in test case.

tests/core/test_runner.py
reata, line 1: Make sure we test against both GraphOperator(s).

reata (changes requested):
Last time I really did't know how to do a subset of lineage when building the graph. Now it's clear you're doing it at graph reading time. It makes sense. I posted a few graph-based considerations for discussion, but I'm good with the direction.

Also I'd like to know how much performance gain it creates. It doesn't have to be as detailed as a full-fledged benchmark that you did before. Just some rough number in production workload would be good.

Thanks again for this POC, amazing work!

sqllineage/core/holders.py
rubytobi: I've opted for an approach where the caller can choose how to filter for the column/node, rather than us being prescriptive about it, i.e. the caller supplements a callable, e.g. a lambda expression.

```python
runner = LineageRunner(sql)

# find_nodes returns every Column/Table/Path vertex matching the predicate
candidates = runner.find_nodes(
  lambda v: isinstance(v, Column) and v.raw_name == "col1"
)

# caller resolves ambiguity (e.g. picking one, or erroring if len != 1)
col = next(c for c in candidates if isinstance(c, Column))

lineage = runner.get_column_lineage(node=col)
```

We don't need the CLI, that was more as a demonstration and example.

rubytobi:
**Benchmark on our internal example using Athena**
For the filtered columns performance, I've taken 10 random Column nodes.

* Small graph (70 nodes, 140 edges)
  - networkx: unfiltered call takes 0.021s; node-filtered calls take 0.0005-0.0015s, averaging ~4% of the unfiltered cost (3-7% range).
  - rustworkx: unfiltered call is already sub-millisecond (0.00072s), so node-filtered calls (0.00016-0.00048s) average ~35% of that (22-67%). The percentage looks worse than networkx's, but that's an artifact of both numbers being tiny, at these absolute times, measurement noise dominates the ratio.

* Large graph (11,181 nodes, 22,371 edges)
  - networkx: *still iterating over countless nodes while I'm writing this* 😉
  - rustworkx: unfiltered call takes 80.7 seconds; node-filtered calls: 0.038-0.14 seconds, averaging ~0.07% of the unfiltered cost (0.05-0.18%). 

On realistic large graphs, node-filtering is blazing fast. It drops from 80-seconds plus post-filtering for us to a sub-150ms one. On tiny/small graphs, the benefit is negligible-to-moderate.

tests/core/test_runner.py
rubytobi: I've added more tests to cover both graph operators.

sqllineage/core/holders.py
rubytobi: I've added `GraphOperator.ancestors`/`descendants` using the native networkx and rustworkx methods over lineage edges, and added consistent errors when nodes don't exists using `KeyError`.

sqllineage/core/holders.py
rubytobi: There is actually a secondary issue which I've aligned so that both graph operators demonstrate the same behaviour:
```shell
--- no self-loop edge ---
networkx:  [['A']]
rustworkx: []

--- with a real self-loop edge ---
networkx:  [['A']]
rustworkx: [[0, 0]]
```

[self_loop_inconsistency.py](https://github.com/user-attachments/files/31223848/self_loop_inconsistency.py)

sqllineage/runner.py
```diff
@@ -132,46 +135,73 @@ def statements(self) -> list[str]:
...
+    def intermediate_tables(self) -> list[Table | Path]:
         """
         a list of intermediate :class:`sqllineage.models.Table`
         """
         return sorted(self._sql_holder.intermediate_tables, key=lambda x: str(x))
 
+    @lazy_method
+    def find_nodes(
+        self, predicate: Callable[[Column | Table | Path], bool] | None = None
+    ) -> list[Column | Table | Path]:
```
reata, line 161: It's a bit counter-intuitive for `lr.find_nodes()` to return `[]`. I don't see much value added to allow `predicate` to default to None. Why don't we make it mandatory?

reata (commented):
We're almost there. Allow me a few more days to review the details. 

Meanwhile, posting one comment for consideration.

sqllineage/runner.py
rubytobi: It all comes down to style, I don't mind empty lists but happy to coerce to `None`. And the optional predicate could also become just an always `True` callable. I'll simplify both.

reata: If user is giving us a lambda predicate that we indeed couldn't find any matches in the graph, returning an empty list makes sense.

However, if the predicate is None, returning either all the nodes in the graph or empty list has its own justification. I say we should prohibit user from passing a None predicate, always ask for a Callable. No default None or always True lambda. (If user explicitly provides always True lambda, that's his choice, our no default logic handles it fine)

sqllineage/core/holders.py
```diff
@@ -35,23 +57,50 @@ def get_column_lineage(
...
+            if exclude_subquery_columns:
+                path = [n for n in path if not isinstance(n.parent, SubQuery)]
+                if len(path) <= 1:
+                    continue
+            columns.add(tuple(path))
         return columns
 
+    def find_nodes(
+        self, predicate: Callable[[Column | Table | Path], bool]
+    ) -> list[Column | Table | Path] | None:
```
reata: Just return `list[Column | Table | Path]` is fine

sqllineage/core/holders.py
```diff
@@ -35,23 +57,50 @@ def get_column_lineage(
...
+        :class:`sqllineage.core.models.Column` instances, e.g.
+        ``[v for v in find_nodes(predicate) if isinstance(v, Column)]``.
+        """
+        return [
+            v
+            for v in self.go.retrieve_vertices_by_props()
+            if predicate is not None
+            and isinstance(v, (Column, *DATASET_CLASSES))
+            and predicate(v)
+        ] or None
```
reata: No need for the None fallback

sqllineage/runner.py
```diff
@@ -132,46 +135,76 @@ def statements(self) -> list[str]:
...
+    def intermediate_tables(self) -> list[Table | Path]:
         """
         a list of intermediate :class:`sqllineage.models.Table`
         """
         return sorted(self._sql_holder.intermediate_tables, key=lambda x: str(x))
 
+    @lazy_method
+    def find_nodes(
+        self, predicate: Callable[[Column | Table | Path], bool]
+    ) -> list[Column | Table | Path] | None:
```
reata: Same here, once holders.py is updated, we can just return `list[Column | Table | Path]`

sqllineage/core/graph/utils.py
```diff
@@ -0,0 +1,66 @@
...
+    true_targets = descendants & targets
+    if seed in sources:
+        true_sources.add(seed)
+    if seed in targets:
+        true_targets.add(seed)
+    if not true_sources or not true_targets:
+        return []
+
+    up_graph = graph.get_sub_graph(*(ancestors | {seed}))
+    down_graph = graph.get_sub_graph(*(descendants | {seed}))
```
reata: We can just build one cone_graph here, and use the same cone_graph to derive up_paths and down_paths.

sqllineage/core/graph/utils.py
```diff
@@ -0,0 +1,66 @@
...
+    self_paths = (
+        up_graph.list_lineage_paths(seed, seed)
+        if seed in true_sources and seed in true_targets
+        else []
+    )
+
+    up_paths = [[seed]] if seed in true_sources else []
+    up_paths += list_lineage_paths_between(up_graph, true_sources - {seed}, {seed})
+    down_paths = [[seed]] if seed in true_targets else []
+    down_paths += list_lineage_paths_between(down_graph, {seed}, true_targets - {seed})
```
reata: Once cone_graph is unified, it's worth a comment to explain why calculate up_paths and down_paths separately and later stitching them together is preferred over iterating the whole cone_graph for performance considerations.

sqllineage/core/graph/rustworkx.py
```diff
@@ -211,11 +223,14 @@ def merge(self, other: GraphOperator) -> None:
             )
 
     def list_lineage_paths(self, src_vertex: Any, tgt_vertex: Any) -> list[list[Any]]:
```
reata, line 225: When the graph is cyclic, rustworkx return path as cyclic like `[v2,v1,v5,v3,v4,v2]`, not just self-loop edge, this is now inconsistent with NetworkxGraphOperator and violates the `list_lineage_paths` contract. We need the same `if src_vertex == tgt_vertex:` branch as networkx side.

sqllineage/runner.py
```diff
@@ -32,6 +33,8 @@ def lazy_property(func):
 
 
 class LineageRunner:
+    _sql_holder: SQLLineageHolder  # For mypy attribute checking; set by _eval()
+
```
reata: Instead of adding an explicit type for _sql_holder, we can set _eval to return None to so mypy can inference the type of _sql_holder automatically.

sqllineage/core/holders.py
```diff
@@ -35,23 +57,50 @@ def get_column_lineage(
...
+        source_column_set = set(source_columns)
+        target_column_set = set(target_columns)
+
+        if node is None:
+            raw_paths = list_lineage_paths_between(
+                column_graph, source_column_set, target_column_set
+            )
+        else:
+            if node not in all_column_set:
+                return set()
```
reata: We can short circuit early for node not in graph scenario to save column_graph calculation.

tests/core/test_runner.py
```diff
@@ -42,6 +48,223 @@ def test_get_column_lineage_exclude_subquery_inpath():
...
+        "insert into tabX select col1 from tab0; "
+        "insert into tabSeed select col1 from tabX; "
+        "insert into tabX select col1 from tabSeed; "
+        "insert into tabY select col1 from tabX"
+    )
+    seed = _col("col1", "tabSeed")
+    with SQLLineageConfig(GRAPH_OPERATOR_CLASS=graph_operator):
+        result = LineageRunner(sql).get_column_lineage(node=seed)
+        for path in result:
+            assert len(path) == len(set(path)), path
```
reata: since result is empty, it never goes into the loop. We might as well just `assert LineageRunner(sql).get_column_lineage(node=seed) == []`

tests/core/test_runner.py
```diff
@@ -42,6 +48,223 @@ def test_get_column_lineage_exclude_subquery_inpath():
...
+# one column per table, so a bare table name resolves to exactly one column
+_SINGLE_COL_SQL = (
+    "insert into tab2 select col1 from tab1; insert into tab3 select col1 from tab2"
+)
+
+
+def _col(name: str, table: str) -> Column:
+    col = Column(name)
+    col.parent = Table(table)
+    return col
```
reata: I'd suggest we refactor this function as `def build_column(col: ColumnQualifierTuple) -> Column:`, move it to `tests/helpers.py`. `_assert_column_lineage` function can reuse this same utility there, making the test suites for consistent.

In `test_runner.py`, it becomes `build_column(ColumnQualifierTuple("col1", "tab2"))`. This is more verbose, but worth the change given the consistency.

tests/core/test_runner.py
```diff
@@ -42,6 +48,223 @@ def test_get_column_lineage_exclude_subquery_inpath():
...
+
+
+def _naive_filter(all_paths, node):
+    """reference implementation: compute everything, keep paths touching node"""
+    return {path for path in all_paths if node in path}
```
reata: This is just used once, let's not create a function for now.

reata (changes requested):
I did a final in-depth review, and attached a git patch for your reference. Do let me know if there's anything that doesn't make sense.

[pr803-review.patch](https://github.com/user-attachments/files/31610497/pr803-review.patch)

rubytobi:
Thanks @reata for reviewing this. As always, appreciate your thoroughness and ongoing maintenance of this great package!

reata (approved):
LGTM. Huge thanks for the quick turnaround @rubytobi

PR #822, opened 2026-08-25
Add wraps decorator to lazy_method function

rubytobi:
`lazy_method`/`lazy_property` don't preserve the wrapped function's `__name__`/`__doc__`, which breaks Sphinx autodoc and IDE tooltips for every decorated method. This has been the case since the decorators were introduced. Adding `functools.wraps` is a one-line, behavior-preserving fix which improves introspection and IDE support.

### DIFF
--- a/sqllineage/runner.py
+++ b/sqllineage/runner.py
@@ -1,6 +1,7 @@
 import logging
 import warnings
 from collections import OrderedDict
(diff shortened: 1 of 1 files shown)


reata (approved):
Thanks. Clean change as always.

PR #838, opened 2026-09-16
fix: expand wildcard selected from CTE

### LINKED ISSUE #630: SELECT * from CTE fails to resolve column lineage
**Describe the bug**
Column lineage fails to resolve when referencing a SELECT * from a CTE.

Not sure if this is the same issue as https://github.com/reata/sqllineage/issues/303 as it looks similar just 1 layer deeper with a secondary select * CTE.

Below is a simplified example, our production scenarios are more complicated but it boils down to this.

**SQL**
```sql
CREATE TABLE MAIN.FOOBAR AS (
    WITH FOO AS (
        SELECT COL1, COL2 FROM FROM MAIN.BAR
    )
    SELECT * FROM FOO
)
```

Column Lineage fails to resolve columns defined in FOO

**To Reproduce**
```python
from sqllineage.runner import LineageRunner
sql1 = """
CREATE TABLE MAIN.FOOBAR AS (
    WITH FOO AS (
        SELECT COL1, COL2 FROM FROM MAIN.BAR
    )
    SELECT * FROM FOO
)
"""
LineageRunner(sql1).print_column_lineage()
```
actual output is
```
main.foobar.* <- foo.*
```
I'd expect it to return this instead
```
main.foobar.col1 <- main.bar.col1
main.foobar.col2 <- main.bar.col2
```

 - 3.11.5

 - 1.5.3

We noticed the issue using the snowflake dialect for our own production use cases

manan28025:
Closes #630

`SELECT *` from a CTE was left as `cte.*` in column lineage, while the same query written with a subquery resolved to real columns.

`CteExtractor` extracts the CTE queries after the statement that selects from them, so when `SelectExtractor` expands the wildcard the CTE has no known columns yet. Subqueries don't have this problem because they're extracted before the outer query expands its wildcards.

Changes:
- `CteExtractor` expands wildcards after the CTE queries are merged, for each CTE in definition order and then for the target table, so a CTE can also select `*` from an earlier CTE.
- `SelectExtractor` no longer expands wildcards selected from a CTE, since the CTE's columns aren't known there.
- `expand_wildcard` only expands from a subquery whose columns are fully known. A column can be attached to a subquery just by being referenced, and expanding from that partial set dropped the rest of the lineage. This keeps `test_nested_cte_with_partial_wildcard_expansion` passing and fixes the same case for nested subqueries, where master reports only `t2.col1 -> tab1.col1`.

A `SELECT *` from an inline subquery that itself selects `*` from a CTE is still left unexpanded.

Testing: new tests cover `*` and `alias.*` from a CTE, an explicit column plus `*`, UNION of CTEs, a CTE selecting `*` from an earlier CTE, CTAS with a CTE, a CTE over a table wildcard with metadata, and the nested subquery case. They fail on master. `black --check`, `ruff check`, `mypy` and the full test suite pass locally on Python 3.13, except `tests/core/test_drawing.py::test_handler`, which fails the same way on master in my environment.


reata:
Thanks for your contribution. The fix is clean and well-scoped.

My long term plan is to introduce a standalone resolver, decoupled from extraction. Wildcard expansion currently lives on the holder and is triggered from three places with three different timing strategies: `select.py` defers CTE sources via `resolve_cte=False`, `cte.py` re-expands in definition order, and the sqlparse analyzer expands after its subquery recursion (cross-statement resolution adds a fourth path through `session.register_session_metadata` in `LineageRunner._eval`). This deeply couples wildcard expansion to lineage analyzer, and we might meet future corner case like this.

manan28025:
Thanks for merging, and for the context on the resolver plan. Makes sense — happy to help there if useful once you have a design.

reata:
> Thanks for merging, and for the context on the resolver plan. Makes sense — happy to help there if useful once you have a design.

Sure. I'll create a issue with design later and ping you to see if you're interested. Thanks for helping.

PR #839, opened 2026-09-16
fix: keep schema of tables read in scalar subquery

### LINKED ISSUE #634: When parsing a segment of SQL, although the field-level relationships exist, the source table information is lost.
**Describe the bug**
When parsing a segment of SQL, although the field-level relationships exist, the source table information is lost.
 existing table prefixes have also been replaced with '<default>'.

**SQL**
Paste the SQL text here. For example:
```sql=
insert into foo.tb1  select f1,f2,(select f3 from  bar.tb1) b3 from  foo.tb2
```

For example:
```pycharm  run 

MetaData = {"bar.tb1": ["f1", "f2", "f3"]}

provider = DummyMetaDataProvider(MetaData)

LineageRunner = sqllineage.runner.LineageRunner(sql, metadata_provider=provider, verbose=True)

LineageRunner.print_column_lineage()
print("=======source_tables:")
print(LineageRunner.source_tables)
```

result:
```
foo.tb1.b3 <- <default>.tb1.f3
foo.tb1.f1 <- foo.tb2.f1
foo.tb1.f2 <- foo.tb2.f2
=======source_tables:
[Table: foo.tb2]
```

**Expected behavior**
I hope to correctly display the prefix **bar. for  tb1.f3**    and **bar.tb1 in source_tables**

 - 3.8.17

 - 1.5.3

**Additional context**

manan28025:
Closes #634

For a scalar subquery in the SELECT list, the column lineage lost the schema of the table it reads from:

```sql
insert into foo.tb1 select f1, f2, (select f3 from bar.tb1) b3 from foo.tb2
```

gave `foo.tb1.b3 <- <default>.tb1.f3`. If the outer query reads a table with the same name, the column was attributed to the wrong table: `(select f3 from bar.tb2) ... from foo.tb2` gave `foo.tb2.f3`.

The subquery's source columns were qualified with `src_col.parent.raw_name`, which drops the schema, and `Column.to_source_columns` then resolves that bare name against the outer query's tables. They're now qualified with the full table name, which resolves to exactly that table. Columns whose parent isn't a table are skipped, since they can't be referenced from the outer query and previously raised `AttributeError: 'SubQuery' object has no attribute 'raw_name'` (for example `(select x from (select 1 as x) s)`).

The same code exists for both the sqlfluff and sqlparse parsers, so both are changed.

Testing: added tests for a schema-qualified scalar subquery (with and without an alias), a same-named table in the outer query, and a literal read from a nested subquery. They fail on master and pass with the fix, on both parsers. `black --check`, `ruff check`, `mypy` and the full test suite pass locally, except `tests/core/test_drawing.py::test_handler`, which fails the same way on master in my environment.


reata (changes requested):
The code change is quite straight-forward. 

Please resolve the conflict and it's good to merge.

PR #840, opened 2026-09-18
fix: handle EXCEPT/REPLACE in wildcard column lineage (BigQuery)

### LINKED ISSUE #576: Column Level Lineage: SELECT * EXCEPT() not showing all columns (BigQuery)
**Describe the bug**
* When using SELECT * EXCEPT(), the column level lineage is not detailed.

**SQL**
```sql
CREATE OR REPLACE TEMP TABLE `table_1` AS

SELECT
    * EXCEPT(B)
FROM `dataset.table_0`
;
```

**To Reproduce**
*Note here we refer to SQL provided in prior step as stored in a file named `test.sql`*
```python
import json
from sqllineage.core.metadata.dummy import DummyMetaDataProvider
from sqllineage.runner import LineageRunner

with open("test.sql") as f:
    sp = f.read()

with open('metadata.json') as file:
    metadata = json.load(file)
provider = DummyMetaDataProvider(metadata)

lineage = LineageRunner(sp, dialect="bigquery", metadata_provider=provider, verbose=True)
lineage.print_column_lineage()
```
Metadata.json:
```
{
    "dataset.table_0": [
        "A",
        "B",
        "C",
        "D"
      ]
}
```

**Result**
```
<default>.table_1.* except(b) <- dataset.table_0.* except(b)
```

**Expected behavior**
```
<default>.table_1.A <- dataset.table_0.A
<default>.table_1.C <- dataset.table_0.C
<default>.table_1.D <- dataset.table_0.D
```

 - 3.9.12

 - 1.5.1

manan28025:
Closes #576

`SELECT * EXCEPT(B) FROM t` produced a single bogus column edge named `* except(b)` instead of expanding to the real columns minus B:

```
<default>.table_1.* except(b) <- dataset.table_0.* except(b)
```

`extract_column_qualifier` derived the wildcard's column name from the entire `wildcard_expression` segment's raw text, which includes the EXCEPT/REPLACE clause, not just the `*`. The same bug affects `* REPLACE(...)`.

Changes:
- `extract_column_qualifier` now derives the column/qualifier from the `wildcard_identifier` child only (e.g. "*" or "t.*"), ignoring EXCEPT/REPLACE.
- A new `extract_wildcard_except_columns` reads the EXCEPT'd column names from the `select_except_clause`.
- `Column` gained an `except_columns` attribute, set on the wildcard's target Column when built.
- `_replace_wildcard` filters those names out of the source columns before expanding, so EXCEPT'd columns are correctly dropped from the lineage.
- `REPLACE(expr AS col)` doesn't remove a column, so once the wildcard is parsed correctly it already expands to all columns unchanged; no separate handling needed.

This only touches the sqlfluff-based parser: EXCEPT/REPLACE is BigQuery/DuckDB-specific syntax the legacy sqlparse-based parser doesn't understand at all (it silently keeps every column today), so new tests set `test_sqlparse=False`.

Testing: 11 new tests cover a single EXCEPT'd column, multiple, a qualified wildcard (`t.* EXCEPT(...)`), REPLACE alone, EXCEPT combined with REPLACE, and EXCEPT when the source is a subquery rather than a metadata provider. They fail on master and pass with the fix. `black --check`, `ruff check`, `mypy` and the full test suite pass locally, except `tests/core/test_drawing.py::test_handler`, which fails the same way on master in my environment.


sqllineage/core/parser/sqlfluff/models.py
```diff
@@ -131,12 +132,14 @@ def of(column: BaseSegment, **kwargs) -> Column:
...
             if source_columns:
                 column_name = None
+                except_columns: list[str] = []
                 for sub_segment in list_child_segments(column):
                     if sub_segment.type == "column_reference" or is_wildcard(
                         sub_segment
                     ):
                         if cqt := extract_column_qualifier(sub_segment):
                             column_name = cqt.column
+                        except_columns = extract_wildcard_except_columns(sub_segment)
```
reata, line 142: Current code is correct in behavior because `extract_wildcard_except_columns` also works for non wildcard.

But please insert a `if is_wildcard(sub_segment)` before so that the code is more readable.

tests/sql/column/test_metadata_wildcard.py
reata, line 1: 1. The newly added unit tests are all dialect specific, put them in a new `tests/sql/column/test_metadata_wildcard_dialect_specific.py` file. You can refer to other `_dialect_specific.py` files as example.
2. use conftest.py defined schema to construct tests instead of leveraging `generate_metadata_providers`. A few unit tests have to use it for reasons documented in comment. That doesn't apply to these new test cases.
3. As you also point out, there're other SQL dialect which also supports `select * except(...)` syntax. We should test all that are applicable. Note duckdb is not one of them, check below. I'm mostly talking about clickhouse and maybe databricks/sparksql
4. A syntax variant of `select * exclude(...)` exists for dialect like duckdb, snowflake. It's a nice-to-have to support it. Since supporting it may require further change on `extract_wildcard_except_columns` util, I'm fine if we want to defer.

reata (changes requested):
The core logic change is very clear. A few recommendations on organization of test files to follow the convention of this repo.

